PR #5, opened 2018-03-23
fix(test): fix skipped tests, closes #4

gq0:
requires this https://github.com/grid-x/images/pull/9

PR #6, opened 2018-03-23
fix/skip_linting_2

gq0:
Should be merged after https://github.com/grid-x/modbus/pull/5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 GO_FILES := $(shell find . -type f -name "*.go")
 GO_BUILD := CGO_ENABLED=0 go build -ldflags "-w -s"
-#GO_TOOLS := gridx/golang-tools:master-839443d
-GO_TOOLS := gridx/golang-tools:1.8.7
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -10,15 +10,17 @@ all: bin/
...
 test:
-	go test -v $(shell glide nv)
+	diagslave -m tcp -p 5020 & go test -run TCP -v $(shell glide nv)
+	socat -d -d pty,raw,echo=0 pty,raw,echo=0 & diagslave -m rtu /dev/pts/1 & go test -run RTU -v $(shell glide nv)
+	socat -d -d pty,raw,echo=0 pty,raw,echo=0 & diagslave -m ascii /dev/pts/3 & go test -run ASCII -v $(shell glide nv)
 
 .PHONY: lint
 lint:
-	golint -set_exit_status $(shell glide nv)
+	golint -set_exit_status
```
jhedev: wouldn't this exclude subpackages from liniting?

gq0: Yeah, the only sub package is the testing one and it was too much effort to fix it too.

PR #7, opened 2018-10-17
Provide generic logging interface

andig:
Migrated from https://github.com/goburrow/modbus/pull/33

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,11 @@ import (
 	"fmt"
 )
 
+// logger is the interface to the required logging functions
+type logger interface {
+	Printf(format string, v ...interface{})
+}
+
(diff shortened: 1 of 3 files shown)


gq0:
Hi @andig, thanks for forwarding the PR. 

Just to understand your intend first. You want to register the logrus Logger implementation at the library, correct? Would the following code snipped already be sufficient for your purpose?

```golang
handler := modbus.NewTCPClientHandler(address)
handler.Logger = log.New(logrus.New().Writer(), "your prefix ", log.LstdFlags) 
```

andig:
Yes it would. Or I could inject any other object that implements Printf. Without changing anything really its already decoupled from golang log implementation.

PR #10, opened 2018-12-02
Align log messages

andig:
Improve log output my aligning output strings visually

### DIFF
--- a/asciiclient.go
+++ b/asciiclient.go
@@ -178,7 +178,7 @@ func (mb *asciiSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, e
 	mb.serialPort.startCloseTimer()
 
 	// Send the request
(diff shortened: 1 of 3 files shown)


asciiclient.go
```diff
@@ -178,7 +178,7 @@ func (mb *asciiSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, e
 	mb.serialPort.startCloseTimer()
 
 	// Send the request
-	mb.serialPort.logf("modbus: sending %q\n", aduRequest)
+	mb.serialPort.logf("modbus: send %0 x\n", aduRequest)
```
gq0: I think the 0 padding can be skipped here, or?

andig: ahh, missed one, fixed

PR #12, opened 2019-01-09
fix(rtuclient): read data incrementally byte for byte

jhedev:
also adds tests for this

still testing this fix ...


rtuclient.go
```diff
@@ -129,44 +203,16 @@ func (mb *rtuSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, err
...
-	}
-
-	if err != nil {
-		return
-	}
-	aduResponse = data[:n]
-	mb.serialPort.logf("modbus: recv % x\n", aduResponse)
+	mb.mu.Lock()
+	defer mb.mu.Unlock()
+	data, err := readIncrementally(aduRequest[0], aduRequest[1], mb.port)
```
gq0: I think we need a check here to confirm that we have at least one full RTU frame, otherwise `readIncrementally` won't terminate.

rtuclient.go
```diff
@@ -115,6 +123,72 @@ type rtuSerialTransporter struct {
...
+func readIncrementally(slaveID, functionCode byte, r io.Reader) ([]byte, error) {
+	n := 0
+	data := make([]byte, rtuMaxSize)
+
+	state := stateSlaveID
+	var length, toRead byte
+	crcCount := 0
+
+	for {
+		if r == nil {
```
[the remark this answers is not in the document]
jhedev: good catch. nah it should probably return right away with an error

rtuclient.go
jhedev: But `mb.port` points directly to the serial port. And `ReadAtLeast`  will simply block, I think. So it should be fine to just wait for `readIncrementally` to return.

What do you think?

Also if we read the data into a buffer and pass it to readIncrementally only if there is at least a full RTU frame the benefit of readIncrementally is lost because if there is garbage at the beginning the data won't contain a full RTU Frame

jhedev: And btw: this works totally fine for some time in one of our customers setups where we have frequent errors and missing data.

jhedev:
@gq0 please have a look again. I update the method to cover all function codes. It now also passes the tests :)

rtuclient.go
gq0: All right :+1:

PR #14, opened 2019-08-11
readIncrementally buffer overflow panic and infinite loop

MattBrittan:
Not sure if you are accepting pull requests on this fork but it looked like you had some useful enhancements so I thought it was worth sending this.

1. readIncrementally could cause a panic due to an overflow in the buffer 'data'. This would only happen where a large value was received for 'length' and then sufficient data was sent through to oveflow the buffer (I had this occur when testing a device with dodgy cabling).

2. It was possible that readIncrementally would never exit. This would occur if the serial line delivered data continuously without ever delivering what was expected. Again this happened to me when testing with dodgy cabling (not the end of the world but made debugging confusing).

### DIFF
--- a/rtuclient.go
+++ b/rtuclient.go
@@ -205,6 +205,11 @@ func readIncrementally(slaveID, functionCode byte, r io.Reader) ([]byte, error) 
 		case stateReadLength:
 			// read length byte
 			length = buf[0]
+			// max length = rtuMaxSize - SlaveID(1) - FunctionCode(1) - length(1) - CRC(2)
+			if length >= rtuMaxSize-5 {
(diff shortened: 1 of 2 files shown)


andig:
Just an idea- would it be more go-ish to use a context with deadline and pass that as first parameter instead of an appended timeout?

MattBrittan:
>Just an idea- would it be more go-ish to use a context with deadline
Did consider that but as it's an internal function I did not think that using a context would have any real benefit but would make the function less readable (less clear on why readIncrementally exits). If Send was to be changed to accept a context (breaking change but something that could be useful) readIncrementally should definitly accept a context.

PR #15, opened 2019-08-14
correct readIncrementally buf length check and add tests

MattBrittan:
Added tests for the max message length check in readIncrementally
Noticed that the length check was rejecting messages that it should not have (by one byte) so corrected that.


MattBrittan:
Found a further issue (device returning length of 0 followed by lots of data led to crash). So closing this request and opening a new one (working with a really bad device at the moment - still guess its a good test of the code!).

PR #16, opened 2019-08-15
correct readIncrementally buf length checks and add tests

MattBrittan:
Added tests for the max message length check in readIncrementally
Fixed possible panic if device returned length of 0 (followed by lots of data)
Noticed that the length check was rejecting messages that it should not have (by one byte) so corrected that.


rtuclient.go
```diff
@@ -139,6 +139,17 @@ type rtuSerialTransporter struct {
 	serialPort
 }
 
+// invalidLengthError is returned by readIncrementally when the modbus response would overflow buffer
+// implemented to simplify testing
+type invalidLengthError struct {
```
jhedev: wouldn't it make sense to export this error? So we can check for it outside the pkg?

MattBrittan: I saw it as an implementation detail (it was only added to simplify testing) and made it private to avoid committing to retaining it in the future.

The error should be very rare (it would only really be expected to happen when corrupt data is supplied by the device). Happy to export it if that is preferred but I cannot foresee a circumstance where a user would want to check for this specific error (the timeout seems like a more likely candidate for that).

gq0: Because it is a library and user may handle this corrupt data case differently (e.g. retry or terminate) I would also suggest to export it.

MattBrittan: As it appears that exporting this is the preferred approach I have made the change.

PR #19, opened 2019-11-06
feat(rtu): detach serial port form activity tracking

frzifus:
Until now the serial port has taken over many tasks. It was responsible for
timeouts, new connections and closeing the connection. This makes the port quite
difficult to replace. Now the port has been replaced by an interface.
The activity tracking is done on a upper level.

Signed-off-by: Benedikt Bongartz <b.bongartz@gridx.de>

### DIFF
--- a/rtuclient.go
+++ b/rtuclient.go
@@ -8,7 +8,10 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"sync"
 	"time"
+
+	"github.com/grid-x/serial"
 )
 
 const (
@@ -46,6 +49,15 @@ const (
 type RTUClientHandler struct {
 	rtuPackager
 	rtuSerialTransporter
+}
+
+// NewRTUClientHandlerWithPort uses the RTUClientHandler on top of the given port
+func NewRTUClientHandlerWithPort(p Port) *RTUClientHandler {
+	handler := &RTUClientHandler{}
+	handler.Port = p
+	handler.Timeout = serialTimeout
+	handler.IdleTimeout = serialIdleTimeout
+	return handler
 }
 
 // NewRTUClientHandler allocates and initializes a RTUClientHandler.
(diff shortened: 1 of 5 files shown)


rtuclient.go
```diff
@@ -251,27 +270,27 @@ func readIncrementally(slaveID, functionCode byte, r io.Reader, deadline time.Ti
...
 	//function := aduRequest[1]
 	//functionFail := aduRequest[1] & 0x80
 	bytesToRead := calculateResponseLength(aduRequest)
 	time.Sleep(mb.calculateDelay(len(aduRequest) + bytesToRead))
 
-	mb.mu.Lock()
-	defer mb.mu.Unlock()
-	data, err := readIncrementally(aduRequest[0], aduRequest[1], mb.port, time.Now().Add(mb.serialPort.Config.Timeout))
-	mb.serialPort.logf("modbus: recv % x\n", data[:])
+	mb.Port.Lock()
```
guelfey: shouldn't we lock this before the write already?

serial.go
```diff
@@ -20,82 +18,7 @@ const (
 
 // serialPort has configuration and I/O controller.
 type serialPort struct {
```
guelfey: do we need this at all now? thought we only have the `Port` interface which is implemented by `serial.Port`

rtuclient.go
frzifus: y, would make sense :100:

serial.go
frzifus: `Port` extends `serial.Port` by `sync.Locker`. We decided to handle op locks in this package.
SEE: https://github.com/grid-x/serial/pull/2

serial_test.go
```diff
@@ -22,15 +24,15 @@ func TestSerialCloseIdle(t *testing.T) {
...
+	s := rtuActivityTracker{
+		Port:        port,
 		IdleTimeout: 100 * time.Millisecond,
 	}
 	s.lastActivity = time.Now()
 	s.startCloseTimer()
 
 	time.Sleep(150 * time.Millisecond)
-	if !port.closed || s.port != nil {
+	if !port.closed && s.Port == nil {
```
gq0: Why changing the logic here?

jhedev (commented):
I guess I'm not the right one to review this

RTTTC:
I guess I could at least bump this up? :)

PR #22, opened 2019-12-24
Always reconnect if connection was closed

andig:
This PR is a minor simplification of the reconnection logic.


andig:
> Basically we follow the angular guide.

That's surprising as literally none of the commits in this repo are following this guide. Hope I got it right anyway. Personally, I have switched to stashed commits which allows to adjust the commit message as desired.

PR #23, opened 2019-12-26
Switch to go modules

andig:
This PR implements go modules. The library import path represents a `v1` version and remains stable.

This should work for go >= 1.11. Additional changes might be necessary for building/testing on go 1.11 (e.g. enabling `GO111MODULES`). I've not touched these as I cannot look into the test pipeline.

Happy to add any changes requested if this is desired.


andig:
Modules are supported since 1.11. no more need for vendored deps.

go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/grid-x/modbus
+
+go 1.13
```
gq0: Please ensure that it works with go 1.11

go.mod
```diff
@@ -0,0 +1,5 @@
...
+
+require github.com/grid-x/serial v0.0.0-20191104121038-e24bc9bf6f08
```
gq0: This is the only dependency for now. But please keep the vendor directory to ensure that we can run a reproducible build even if the dependencies location or location structure changes.

A while back, I heard about go modules and reproducible builds through dependency servers. What is the state on this?

andig: Afaiks the reproducible build dicsussion is about ensuring that the module depended on is still available in the right version at the time of build. As for the serial module that is under your control, so no problem. In the general case this is addressed by the go proxy. See https://github.com/golang/go/wiki/Modules for details.

> But please keep the vendor directory to ensure that we can run a reproducible build even if the dependencies location or location structure changes.

If you want to do that I wouldn't move to modules as I feel that maintaining 2 dependency mechanisms is more error-prone than a single well-defined one.

That being said it seems that many major projects are moving to go modules and dropping vendor folders. The ones I've seen that have not done so are impacted by improper use of modules in downstream repositories, something that is not a concern here.

Feel free to close if you want to stay with vendored dependency.

guelfey: we could use go modules and still have a vendor directory by using `go mod vendor` and `go build -mod=vendor` which is mainly useful if you depend on non-public repos, but for public repos I don't see a reason why we shouldn't go modules-only

guelfey:
should be enough to set GO111MODULES=on in the build and test targets of the Makefile

go.mod
gq0: At the time of build the public repo could also not be reachable. Github has not a 100% availability :wink: Therefore, I am in favour to keep the vendor directory but move to go modules.

guelfey: yes, but then the build would fail anyway because this repo is also on github :smile:

andig: > yes, but then the build would fail anyway because this repo is also on github 😄

I was just gonna say that. The problem is recursive ;) Anyway, also with go modules, go will still maintain it's local package cache. So once built, modules will be part of local cache irrespective of vendor folder.

Imho it is either or though I cannot ad hoc find any best practice confirming that.

andig: Here is an article regarding go module proxy: https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/

> The Go module proxy is by default caching and storing all the dependencies forever (in immutable storage). This means you don’t have to use any vendor/ folder anymore.

guelfey: > Anyway, also with go modules, go will still maintain it's local package cache.

That's of course useful for local development, but in our CI setup we can't rely on it being up to date or even present

frzifus:
Basically I am a friend of the vendor function. It makes sure that all deps are there. As far as a proxy is involved, I would prefer to run it myself... well, since we only have one dependency and thats a package from us, I would agree to remove the vendor.

andig:
> Basically I am a friend of the vendor function. It makes sure that all deps are there. As far as a proxy is involved, I would prefer to run it myself... well, since we only have one dependency and thats a package from us, I would agree to remove the vendor.

This sounds like a "go". The entire go world has meanwhile switched to modules. I'd be happy to:
- rebase this
- update the minimum go version to something less demanding (1.11)
- upgrade the serial dependency

Would this be sufficient to get modules implemented?

guelfey:
sounds good to me, just keep in mind to stick to `-mod=vendor`

andig:
I still feel that keeping the vendor folder sort of defies the purpose of go modules, but added as requested. The missing example.go seems due to `go mod vendor` removing the examples.

andig:
**UPDATE**: I'm withdrawing this PR as the vendor functionality seems really strange. Not only the examples are gone, it also removes vendor/github.com/grid-x/serial/syscall_windows.go. This might be a go mod issue but I don't have the time for following up. The benefit of adding modules while keeping the vendor folder seems to small to bother.

Thanks for discussing this in the first place!

Issue #11, opened 2019-01-05
TCP error handling


andig:
@gq0 in https://github.com/goburrow/modbus/pull/25#issuecomment-429637421 you were so kind to help with tcp recovery parameters. I've switched to grid-x and am still having some problems getting a stable connection to SE9k. I'm using these parameters:

	handler.Timeout = 300 * time.Millisecond
	handler.ProtocolRecoveryTimeout = 10 * time.Second
	handler.LinkRecoveryTimeout = 60 * time.Second

Still, recovery does only work sometimes:

	17:23:33 modbus: send 18 fc 00 00 00 06 01 03 9c 95 00 02
	17:23:33 modbus: send 18 fc 00 00 00 06 01 03 9c 95 00 02
	17:23:38 modbus: send 18 fc 00 00 00 06 01 03 9c 95 00 02
	17:23:48 Device 1 failed to retrieve opcode 0x9c95, error was: read tcp 192.168.178.36:34614->192.168.178.5:502: i/o timeout
	17:23:48 Device 1 failed to respond - retry attempt 1 of 5
	17:23:48 modbus: send 18 fd 00 00 00 06 01 03 9c 95 00 02
	17:23:48 modbus: send 18 fd 00 00 00 06 01 03 9c 95 00 02
	17:23:48 modbus: send 18 fd 00 00 00 06 01 03 9c 95 00 02
	17:23:48 modbus: send 18 fd 00 00 00 06 01 03 9c 95 00 02
	17:23:48 modbus: send 18 fd 00 00 00 06 01 03 9c 95 00 02
	17:23:48 modbus: recv 18 fd 00 00 00 07 01 03 04 13 87 ff fe
	17:23:48 Device 1 - Frequency: 49.99

Whereas often it does not work:

broken pipe:

	03:46:43 modbus: send 12 f3 00 00 00 06 01 03 9c 8f 00 04
	03:46:43 Device 1 failed to retrieve opcode 0x9c8f, error was: read tcp 192.168.178.36:60398->192.168.178.5:502: read: connection reset by peer
	03:46:43 Device 1 failed to respond - retry attempt 1 of 5
	03:46:43 modbus: send 12 f4 00 00 00 06 01 03 9c 8f 00 04
	03:46:43 Device 1 failed to retrieve opcode 0x9c8f, error was: write tcp 192.168.178.36:60398->192.168.178.5:502: write: broken pipe
	03:46:43 Device 1 failed to respond - retry attempt 2 of 5

[...]

PR #29, opened 2020-07-17
Rewrite link recovery

andig:
Improve #11. It is not really a fix as it doesn‘t handle timeouts due to the filtering of the error types, but it does already apply recovery handling to the write path as well.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -149,6 +149,21 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
 	var data [tcpMaxLength]byte
 	recoveryDeadline := time.Now().Add(mb.IdleTimeout)
 
+	// recover link if deadline permits by closing connection
+	recoverLink := func(err error) bool {
+		res := mb.LinkRecoveryTimeout > 0 &&
+			time.Since(recoveryDeadline) < 0 &&
+			(err == io.EOF || err == io.ErrUnexpectedEOF)
+
+		if res {
+			mb.logf("modbus: close connection and retry, because of %v", err)
+			mb.close()
+			time.Sleep(mb.LinkRecoveryTimeout)
+		}
+
+		return res
+	}
+
 	for {
 		// Establish a new connection if not connected
 		if err = mb.connect(); err != nil {
@@ -158,22 +173,28 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
 		// Set timer to close when idle
 		mb.lastActivity = time.Now()
 		mb.startCloseTimer()
+
 		// Set write and read timeout
-		var timeout time.Time
 		if mb.Timeout > 0 {
-			timeout = mb.lastActivity.Add(mb.Timeout)
+			timeout := mb.lastActivity.Add(mb.Timeout)
+			if err = mb.conn.SetDeadline(timeout); err != nil {
+				return
+			}
 		}
-		if err = mb.conn.SetDeadline(timeout); err != nil {
-			return
-		}
+
 		// Send data
 		mb.logf("modbus: send % x", aduRequest)
 		if _, err = mb.conn.Write(aduRequest); err != nil {
+			if recoverLink(err) {
+				continue
+			}
 			return
 		}
+
 		// Read header first
 		if _, err = io.ReadFull(mb.conn, data[:tcpHeaderSize]); err == nil {
 			aduResponse, err = mb.processResponse(data[:])
+
 			if err == nil {
 				err = verify(aduRequest, aduResponse)
 				if err == nil {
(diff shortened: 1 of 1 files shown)


tcpclient.go
```diff
@@ -158,47 +173,48 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
...
 				err = verify(aduRequest, aduResponse)
 				if err == nil {
 					mb.logf("modbus: recv % x\n", aduResponse)
 					return // everything is OK
 				}
 			}
+
 			if _, ok := err.(ErrTCPHeaderLength); !ok {
-				if mb.ProtocolRecoveryTimeout > 0 && recoveryDeadline.Sub(time.Now()) > 0 {
+				if mb.ProtocolRecoveryTimeout > 0 && time.Since(recoveryDeadline) < 0 {
```
gq0, line 207: From the top of my head I think this requires go 1.11. Maybe let's add a min. go version to the README.md then as well.

tcpclient.go
```diff
@@ -326,8 +342,7 @@ func (mb *tcpTransporter) closeIdle() {
 	if mb.IdleTimeout <= 0 {
 		return
 	}
-	idle := time.Now().Sub(mb.lastActivity)
-	if idle >= mb.IdleTimeout {
+	if idle := time.Since(mb.lastActivity); idle >= mb.IdleTimeout {
```
gq0, line 345: Same es above, requires go 1.11 I think.

tcpclient.go
```diff
@@ -149,6 +149,21 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
 	var data [tcpMaxLength]byte
 	recoveryDeadline := time.Now().Add(mb.IdleTimeout)
 
+	// recover link if deadline permits by closing connection
+	recoverLink := func(err error) bool {
+		res := mb.LinkRecoveryTimeout > 0 &&
+			time.Since(recoveryDeadline) < 0 &&
```
gq0, line 155: Newer go version see below.

tcpclient.go
```diff
@@ -149,6 +149,21 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
...
+			(err == io.EOF || err == io.ErrUnexpectedEOF)
```
gq0, line 156: This would change the behaviour in case of `ErrTCPHeaderLength`

tcpclient.go
```diff
@@ -158,47 +173,48 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
...
+			timeout := mb.lastActivity.Add(mb.Timeout)
+			if err = mb.conn.SetDeadline(timeout); err != nil {
+				return
+			}
 		}
+
 		// Send data
 		mb.logf("modbus: send % x", aduRequest)
 		if _, err = mb.conn.Write(aduRequest); err != nil {
+			if recoverLink(err) {
```
gq0, line 188: Makes sense :+1:

tcpclient.go
```diff
@@ -158,47 +173,48 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
...
-		} else if (err != io.EOF && err != io.ErrUnexpectedEOF) ||
-			mb.LinkRecoveryTimeout == 0 || recoveryDeadline.Sub(time.Now()) < 0 {
-			return
 		}
-		mb.logf("modbus: close connection and retry, because of %v", err)
 
-		mb.close()
-		time.Sleep(mb.LinkRecoveryTimeout)
+		// attempt failed
+		if !recoverLink(err) {
```
gq0, line 215: As stated in the `recoveryLink` function itself. The behaviour is now different one in case of `ErrTCPHeaderLength`

tcpclient.go
guelfey: are you sure? https://golang.org/pkg/time/#Since doesn't list any minimum version

tcpclient.go
guelfey: maybe it's just too early, but I fail to see how the behaviour would change? we still check for ErrTCPHeaderLength before considering a link failure

tcpclient.go
andig: Since is from 2012 (https://github.com/golang/go/commit/20812c490736e18c454c807ed1c65304bd8b7e5d), couldn't find it in the change logs down to 1.8. Should be ok?

tcpclient.go
andig: Same- I agree though that I could make a now.sub->since change globally since it is much easier to read.

tcpclient.go
andig: > maybe it's just too early, but I fail to see how the behaviour would change?

I think so- no change. The only non-network function that is called here is `processResponse` which has a single "own" error type `ErrTCPHeaderLength`.

This is actually why I would propose to remove the limitation to EOF errors in the link recovery function and just always recover the link. Otherwise, the i/o timeout wouldn't currently even be handled.

tcpclient.go
gq0: Maybe I remember wrongly then. If there is no min. the version in the standard lib, then it's safe.

tcpclient.go
gq0: I see, ok, then I was too quick in my judgment, sorry.

gq0:
@frzifus and @guelfey what do you say?

gq0:
@andig as soon as it isn't a draft anymore we can merge it.

andig:
Thank you for the feedback. Not sure my question has slipped through: could we omit the error check from the recoverLink function and *always* close on any error? Alternatively: check at least for

    err := .(net.Error)

and  close on 

    err.IsTimeout()

to handle hung connections? Once that‘s clarified I‘d also address the TcpRtu Handler.

andig:
ping regarding open question

RTTTC:
Any updates on this?

andig:
4 years later. Giving up.

PR #32, opened 2020-08-27
Bug fix: connection stuck forever with mismatched transaction ID

wz2b:
connection) you might go to send the next poll and immediately receive
the previous response.  The result is a mismatch in the transaction ID
from which the connection will never recover.  There are a few ways to
mitigate this, including receiving the response in a loop, but since
this modbus tcp implementation is completely synchronous the simplest
thing to do is to simply flush the receive buffer immediately prior to
sending the next poll.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -155,6 +155,19 @@ func (mb *tcpTransporter) Send(aduRequest []byte) (aduResponse []byte, err error
 			return
 		}
 
+
+		/*
+		 * If an answer to a previously timed-out request is already in teh buffer, this will result
+		 * in a transaction ID mismatch from which we will never recover.  To prevent this, just
+		 * flush any previous reponses before launching the next poll.  That's throwing away
+		 * possibly useful data, but the previous request was already satisfied with a timeout
+		 * error so that probably makes the most sense here.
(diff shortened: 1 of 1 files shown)


wz2b:
Another approach here would be to read the response in a loop.  If the transaction ID is not the one you expected, go around the loop again.  The reason I didn't do that is it's more intrusive of a code change, but if somebody feels strongly that's a better way to solve this let me know.

frzifus:
hey @wz2b thanks for your contribution, i will have a look in the next days. What might help is a unittest that shows the behaviour.

wz2b:
> hey @wz2b thanks for your contribution, i will have a look in the next days. What might help is a unittest that shows the behaviour.

I think that to really test it would require mocking the TCP stream, wouldn't it?  The behavior I would want to see would be send it a mismatched transaction id and see that it closes the connection, fails that transaction, then starts over at transaction id 1 with a new connection.  I'm not really sure how to approach this in go.

guelfey (approved):
Thanks for your contribution :+1: I'll just reformat the comments and commit message to our style

guelfey:
see #36.

PR #33, opened 2020-12-23
feat: add property based tests for encoding/decoding

jhedev:
You can run the individual test with as many checks as possible, e.g.

```
go test -v -run TestTCPEncodeDecode -rapid.checks=10000 # to run 10k checks
```

Currently, everything passes the checks \o/ I introduced small bugs in the logic
of the methods and quickly the tests failed. So they are indeed working :)

Not sure if we want to keep this which is up to the others. IMO it is a nice way
to generate plenty of testcases.


jhedev:
Oh and by default this will run 100 checks (as you can see in the CI output)

rtuclient_prop_test.go
```diff
@@ -0,0 +1,35 @@
...
+		if err != nil {
+			t.Fatalf("error while encoding: %+v", err)
+		}
+
+		dpdu, err := packager.Decode(raw)
+		if err != nil {
+			t.Fatalf("error while decoding: %+v", err)
+		}
+
+		if !cmp.Equal(pdu, dpdu) {
```
gq0, line 31: Do you really need cmp for this basic comparison? I think using std reflect pkg or just directly comparing the `FunctionCode` and `Data` will work. Just trying to minimise dependencies ;)

tcpclient_prop_test.go
```diff
@@ -0,0 +1,36 @@
...
+		if err != nil {
+			t.Fatalf("error while encoding: %+v", err)
+		}
+
+		dpdu, err := packager.Decode(raw)
+		if err != nil {
+			t.Fatalf("error while decoding: %+v", err)
+		}
+
+		if !cmp.Equal(pdu, dpdu) {
```
gq0, line 32: Same as above.

gq0 (changes requested):
Nice. Can you adjust to go.mod change? And I guess cmp is not needed.

rtuclient_prop_test.go
jhedev: Yeah true. For that simple case we could definitely use `reflect`. However, it is more interesting for the line below `cmp.Diff`. However, I can also change that.

gq0: This package is pretty useful. But just for this, it is a little too much 😂 Would be great if you can remove it. Many thanks for the test in general 👍🏻

gq0 (approved):
I guess the vendoring changed inbetween.

frzifus (approved):

andig:
What‘s missing to get this PR merged?

guelfey:
@andig basically just adjusting to the switch to go.mod that we did in the meantime

andig:
@jhedev would you kindly follow up? If not I'll try to take a look soonish.

jhedev:
@andig thanks for bring this back to life :) I'm pretty busy with other stuff right now. but I'll try to find time. If you're faster, feel free to take over.

andig:
@jhedev I think all you need is a `git rebase` plus `go mod tidy`. Tried to do that in https://github.com/grid-x/modbus/pull/41 but without much luck in linking that to your pr.

jhedev:
@andig thanks for the hint. I did the rebase and forced pushed again. 

re `cmp`. I'm not only using the `Compare` but also the `Diff` method. Happy to replace it but I still think it is a useful dependency.

andig:
> Happy to replace it but I still think it is a useful dependency.

I wouldn't mind, especially as it's test-only.

@guelfey good to go?

andig:
@jhedev if `go-cmp` remains a blocker you could use `reflect.DeepEqual` and replace the diff by "expected/got". It's a small helper during testing when an error actually occurs, but nothing absolutely required imho.

guelfey (approved):

PR #35, opened 2020-12-23
feat(vendor): migrate from glide to go modules

frzifus:


andig:
Yes please! I would not set go 1.15 as required though, if I remember the discussion the highest required is 1.13 or lower?

andig:
See https://github.com/grid-x/modbus/pull/23. Lowest version was 1.11.

andig:
Ping @gq0 @frzifus any chance to move this forward? 

Only potential change I'd see is the required version though anybody using go will probably mostly be on lastest anyway.

frzifus:
i guess with the default values 1.13 was needed. so i changed it to that. :)

Carelo (approved):
LGTM!

gq0:
Wasn't it that with version 1.13 go modules language features are introduced by default? I guess if you want to support a version prior you have to set some flags/environment variables.

PR #42, opened 2021-11-22
feat(cmd/modbus-cli): support writing single coil

frzifus:
With this pr the modbus-cli tool is extended by the function `0x05`. This allows to switch individual outputs on and off.

Usage:
```bash
# read
modbus-cli -slaveID=0xFF -address=tcp://127.0.0.1:502 -fn-code=0x01 -type-parse=uint8 -register=54321 -quantity=1 -log-frame

# Turn on
modbus-cli -slaveID=0xFF -address=tcp://127.0.0.1:502 -fn-code=0x05 -type-parse=uint8 -register=54321 -quantity=1 -log-frame -write-value=0xFF00

# Turn off
modbus-cli -slaveID=0xFF -address=tcp://127.0.0.1:502 -fn-code=0x05 -type-parse=uint8 -register=54321 -quantity=1 -log-frame -write-value=0x0000
```

Ref:
https://ozeki.hu/p_5880-mobdbus-function-code-5-write-single-coil.html

Signed-off-by: Benedikt Bongartz benne@klimlive.de

### DIFF
--- a/cmd/modbus-cli/main.go
+++ b/cmd/modbus-cli/main.go
@@ -134,6 +134,20 @@ func exec(
 	switch fnCode {
 	case 0x01:
 		result, err = client.ReadCoils(uint16(register), uint16(quantity))
+	case 0x05:
+		const (
+			coilOn  uint16 = 0xFF00
+			coilOff uint16 = 0x0000
+		)
+		v := uint16(wval)
+		if v != coilOn && v != coilOff {
+			err = fmt.Errorf(
+				"illegal: expect %X to request the output to be on or %X to be off, got: %X",
+				coilOn, coilOff, v,
+			)
+			return
+		}
+		result, err = client.WriteSingleCoil(uint16(register), v)
 	case 0x06:
 		max := float64(math.MaxUint16)
 		if wval > max || wval < 0 {
--- a/cmd/modbus-cli/main.go
+++ b/cmd/modbus-cli/main.go
@@ -139,13 +139,9 @@ func exec(
 			coilOn  uint16 = 0xFF00
 			coilOff uint16 = 0x0000
 		)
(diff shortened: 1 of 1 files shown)


andig:
IMHO the point here is that the on/off values for the coils are non-obvious. Do they depend on modbus implementation's endianess? If not, the on/off values should be part of the api.

guelfey:
I agree, I'd rather not have to know this detail that on is 0xFF00 (I had to look this up right now). Maybe just take any nonzero value as on?

frzifus:
> Do they depend on modbus implementation's endianess?

oh good question, will try to get that out.

> Maybe just take any nonzero value as on?

Sounds fair to me.

andig:
What is missing to get coils support merged?

frzifus:
öhm, seems we're almost celebrating a one year anniversary. :partying_face: :balloon: 

> Do they depend on modbus implementation's endianess?

This is what [modbus.org](https://modbus.org/docs/Modbus_Application_Protocol_V1_1b.pdf) says about it: 
> 4.2 Data Encoding
• MODBUS uses a ‘big-Endian’ representation for addresses and data items. This means that when a numerical quantity larger than a single byte is transmitted, the most significant byte is sent first. 
So for example Register size value 16 - bits 0x1234 the first byte sent is 0x12 then 0x34

Other than that, i haven't found any further info during the last year.

> Maybe just take any nonzero value as on?

Adapted it, now every value that is non-zero represents `on`.

andig:
Recently used https://ozeki.hu/p_5880-mobdbus-function-code-5-write-single-coil.html. Seems we can acceppt any non-zero Value as "on", but devices will only ever accept/return `0xFF00` bigendian.

guelfey (approved):

PR #43, opened 2021-12-23
Use current go idioms

andig:
I'm not 100% sure that that these are supported on go1.13. Couldn't verify on Mac.


frzifus (commented):
Should work with [go 1.13](https://pkg.go.dev/time@go1.13#Until). 

If that were not the case, it should not be a deal breaker to raise the minimum version to at least 1.15. thats at least the version which is used to for the build test in the [ci](https://github.com/grid-x/modbus/blob/master/Makefile#L3) pipe.

guelfey (approved):
Thanks!

PR #44, opened 2021-12-23
Add connect delay

andig:
Some modbus devices from Huawei need a quiet period after initial connection. Since the connect is handled inside the library, the delay can only be added here. If clearer, we could also name this `SilentDelay`. The scope of the changes is minimal, no logic is impacted.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -131,6 +131,8 @@ type tcpTransporter struct {
 	LinkRecoveryTimeout time.Duration
 	// Recovery timeout if the protocol is malformed, e.g. wrong transaction ID
 	ProtocolRecoveryTimeout time.Duration
+	// Silent period after successful connection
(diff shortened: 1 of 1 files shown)


RTTTC (approved):
Looks good!

frzifus (commented):
> Some modbus devices from Huawei need a quiet period after initial connection.

Is this also valid for RTU appliances?

andig:
Only observed on Huawei Tcp, therefore didn‘t apply tu RTU.

gq0 (approved):
Weird quirk. Do you know why Huawei needs this?

andig:
> Weird quirk. Do you know why Huawei needs this?

They have a dongle that connects to the rs485. Just seems crappy :/ Thanks for merging!

PR #47, opened 2022-01-10
fix: print original function code for exceptions

### LINKED ISSUE #46: Modbus errors are misleading
Consider this:

```
[modbus] TRACE 2022/01/07 19:41:45 modbus: send 00 01 00 00 00 06 01 04 13 88 00 01
[modbus] TRACE 2022/01/07 19:41:47 modbus: recv 00 01 00 00 00 06 01 84 04 00 01 00
Power: read failed: modbus: exception '4' (server device failure), function '132'
```

"Function" 132 is actually `84`, which is really `4` plus the error bit. I'd suggest to show the function code minus the error bit in the message here.

guelfey:
Fixes #46.

### DIFF
--- a/modbus.go
+++ b/modbus.go
@@ -89,7 +89,7 @@ func (e *Error) Error() string {
 	default:
 		name = "unknown"
 	}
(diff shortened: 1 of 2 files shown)


gq0 (approved):
Alternatively, displaying it as hex?

PR #49, opened 2022-02-08
fix: retry reading query response on transaction id mismatch

guelfey:
This was previously attempted with the flushAll call, but this doesn't
seem to be reliable in practice - since it sets the read timeout to
time.Now(), this means that in practice the Read in most cases actually
returns directly without attempting to read.

Instead of trying to extend the deadline for the flushing read by some
arbitrary small value, this fix handles the problem at a different
place: if we detect a transaction ID mismatch by 1, we simply ignore the
response and assume that a matching response will follow. This required
some refactoring of the reading code since just retrying the read
without another write wasn't easily possible.

### DIFF
--- a/tcpclient_test.go
+++ b/tcpclient_test.go
@@ -93,6 +93,69 @@ func TestErrTCPHeaderLength_Error(t *testing.T) {
 	_ = ErrTCPHeaderLength(1000).Error()
 }
 
+func TestTCPTransactionMismatchRetry(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ln.Close()
+
+	done := make(chan struct{})
+	defer close(done)
+	data := []byte{0xCA, 0xFE}
+	go func() {
+		conn, err := ln.Accept()
+		if err != nil {
+			t.Error(err)
+			return
+		}
+		defer conn.Close()
+		time.Sleep(1 * time.Second)
+		packager := &tcpPackager{SlaveID: 0}
+		pdu := &ProtocolDataUnit{
+			FunctionCode: FuncCodeReadInputRegisters,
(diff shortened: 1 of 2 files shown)


tcpclient_test.go
```diff
@@ -93,6 +93,69 @@ func TestErrTCPHeaderLength_Error(t *testing.T) {
...
+	defer close(done)
+	data := []byte{0xCA, 0xFE}
+	go func() {
+		conn, err := ln.Accept()
+		if err != nil {
+			t.Error(err)
+			return
+		}
+		defer conn.Close()
+		time.Sleep(1 * time.Second)
```
frzifus, line 113: could this result in flacky test behavior?

frzifus (approved):

Carelo (approved):
Looks good to me, great work :+1:

tcpclient_test.go
Carelo: i think this  should be fine, as long as both requests are sended before this returns.

gq0 (approved):
Awesome work 👍🏻 feel free to merge.
It's actually a behavior change that needs to be guarded with a feature-flag. Could we load two different versions of this lib in our client code base behind a feature flag?

guelfey:
I see it as a bug fix actually - this was supposed to already happen

tcpclient_test.go
guelfey: we have: https://buildkite.com/gridX/modbus/builds/93#ccdff30d-2ec2-47ee-a7cd-3fdd74d17ffb/207

frzifus: But so far no super slow nightly arm build or? which lets all timings exceed :dancers:

gq0:
> I see it as a bug fix actually - this was supposed to already happen

The transaction ID handling seems new to me. But we can also roll it out slowly and see 👍🏻

PR #53, opened 2022-05-23
Validate result length againt request

andig:
In https://github.com/evcc-io/evcc/discussions/3222 we've observed the infamous Huawei inverter return less bytes than requested as "success". Subsequently that leads to downstream `panic` when using the `binary.BigEndian` decoding functions.

Instead, I'd suggest to validate proper length in the library. Same could/should be added for coils/ discrete inputs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -125,6 +125,10 @@ func (mb *client) ReadHoldingRegisters(address, quantity uint16) (results []byte
 		err = fmt.Errorf("modbus: response data size '%v' does not match count '%v'", length, count)
 		return
 	}
+	if count != 2*int(quantity) {
+		err = fmt.Errorf("modbus: response data size '%v' does not match request quantity '%v'", length, quantity)
+		return
+	}
 	results = response.Data[1:]
 	return
 }
@@ -154,6 +158,10 @@ func (mb *client) ReadInputRegisters(address, quantity uint16) (results []byte, 
 	length := len(response.Data) - 1
 	if count != length {
 		err = fmt.Errorf("modbus: response data size '%v' does not match count '%v'", length, count)
+		return
+	}
+	if count != 2*int(quantity) {
+		err = fmt.Errorf("modbus: response data size '%v' does not match request quantity '%v'", length, quantity)
 		return
 	}
 	results = response.Data[1:]


andig:
Friendly ping for review. Just had another incident of this at https://github.com/evcc-io/evcc/issues/8508 which causes application panic. This should be handled in the library, not on the application side.

frzifus (approved):

andig:
@guelfey @gq0 any chance for merging?

andig:
@guelfey @gq0 any chance for merging?

gq0:
I'll consult with one of my colleagues about this matter.

tretmar (commented):
Thanks so much for your contribution here @andig 😊 

Definitely makes sense to double-check if the byte count is matching to the quantity that we've requested 👍 

Could you also add this for 
- ReadCoils
- ReadDiscreteInputs
- ReadWriteMultipleRegisters (maybe)
(We are not using any of those functions, so there is no impact on our side)

_Just my rubber duck debugging brain now_

If byte count is 2 and we also get 2 bytes in the data area, this code would succeed even though we might have requested 4 bytes (quantity = 2).

The higher layers could interpret this in a wrong way as they want to get 4 bytes but instead, they only got 2 bytes.

Yeah I can see that panicing in our application as well 🤔

frzifus:
> Just my rubber duck debugging brain now
>
> If byte count is 2 and we also get 2 bytes in the data area, this code would succeed even though we might have requested 4 bytes (quantity = 2).
> 
> The higher layers could interpret this in a wrong way as they want to get 4 bytes but instead, they only got 2 bytes.
>
> Yeah I can see that panicing in our application as well thinking

Yes, their might be a chance that the higher layer sees this: [panic: runtime error index out of range](https://www.golangprograms.com/panic-runtime-error-index-out-of-range-error-in-golang.html)

andig:
> Could you also add this for

I‘m not familiar with the other functions hence didn‘t add them.

andig:
> If byte count is 2 and we also get 2 bytes in the data area, this code would succeed even though we might have requested 4 bytes (quantity = 2).

Where do you see this in the PR? The comparison is bytes vs 2*words?

tretmar:
> > If byte count is 2 and we also get 2 bytes in the data area, this code would succeed even though we might have requested 4 bytes (quantity = 2).
> 
> Where do you see this in the PR? The comparison is received bytes vs 2*requested words?

My rubber duck brain wrote this for the old code ;) Just to see how it would have behaved. With your code change, this won't succeed ofcourse. Sorry for the confusion!

frzifus:
Good, so whats the state here?

andig:
Ready from by pov unless someone want to add the open points from initial post. Can be done in separate PR.

gq0:
From a protocol specification point of view, it makes sense 👍🏻 
But I fear that not all devices in the field will behave according to specifications. 
If this error is triggered, it will break batch requests on our side, but we will monitor the rollout and merge it now.

andig:
@gq0 as a follow-up we could still make this a sentinel error so it becomes easier to catch?

PR #54, opened 2022-08-29
Wrap open error with device name for context

andig:
Also makes use of `time.Since`. I couldn't find out what Go version `%w` was introduced, so merge with care.

### DIFF
--- a/serial.go
+++ b/serial.go
@@ -5,6 +5,7 @@
 package modbus
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


frzifus (approved):

guelfey (approved):
It was (introduced with Go 1.13)[https://go.dev/doc/go1.13#error_wrapping], which we already have in our go.mod

PR #60, opened 2022-11-06
Check recovery deadline only once

andig:
This PR simplifies checking for `recoveryDeadline` being exceeded.

Note: refactoring the code like this, it seems to indicate that there is something wrong with the usage of `ProtocolRecoveryTimeout`. It's actual value is never used, only checked against being zero. In contrast, `LinkRecoveryTimeout` is used as input to `sleep`.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -218,9 +218,15 @@ func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryD
 					return // everything is OK
 				}
 			}
+
+			// no time left, report error
+			if time.Since(recoveryDeadline) >= 0 {
+				return
+			}
+
 			switch v := err.(type) {
 			case ErrTCPHeaderLength:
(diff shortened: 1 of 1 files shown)


frzifus (commented):
probably worth investing some time in setting up [golangci-lint](https://github.com/golangci/golangci-lint). This way we should be able to find many more of such corners.

cc @guelfey

andig:
Any feedback on this one? Open bug regarding `ProtocolRecoveryTimeout`?

hsteidl (approved):
Good improvement, this simplifies a lot. Especially when looking at corner cases like in https://github.com/grid-x/modbus/pull/66 where the conditions started getting difficult.
This also looks functionality-preserving to me!

hsteidl:
@andig I will merge this PR now as preparation for #66. Another discussion for the `ProtocolRecoveryTimeout` problem would be great as this is still weird

PR #64, opened 2023-03-08
Add support for common register encoding/decoding orders

tretmar:
We often have to use "mid-little-endian" encoding for 32bit registers but the CLI tool didn't offer support for it.
With this PR, I propose to add new CLI flags that can force the order for reading and writing registers.

The supported orders are: "AB", "BA", "ABCD", "DCBA", "BADC", and "CDAB".

I also found out that writing a float32 was not possible as the encoding was wrong.
Thus, I replaced the encoding mechanism here as well.

Note: Since this is now extending the Modbus CLI significantly, I will do a PR afterwards to generate the binaries easier :D

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ test:
 	diagslave -m tcp -p 5020 & diagslave -m enc -p 5021 & go test -run TCP -v $(shell glide nv)
 	socat -d -d pty,raw,echo=0 pty,raw,echo=0 & diagslave -m rtu /dev/pts/1 & go test -run RTU -v $(shell glide nv)
 	socat -d -d pty,raw,echo=0 pty,raw,echo=0 & diagslave -m ascii /dev/pts/3 & go test -run ASCII -v $(shell glide nv)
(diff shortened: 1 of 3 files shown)


andig:
Nice idea. Lets put the decoders somewhere accessible for library consumers, too!

tretmar:
Note: https://github.com/grid-x/modbus/pull/65 needs to be merged before this PR

andig:
We have a ton of encodings in https://github.com/evcc-io/evcc but our naming is odd. Maybe start an entirely new repo for encoding? Would be happy to contribute!

Rania-Zoghlami (approved):
Nice work 👍  lets discuss the encoding/decoding package later ;)

tretmar:
Yes, we are internally working on a separate encoding/decode repo already. We might use it here later :)

andig:
@tretmar feel free to spy on https://github.com/volkszaehler/mbmd/tree/master/encoding and don't forget the strings ;)

tretmar:
Will do that @andig thanks 😊

PR #66, opened 2023-04-05
Fix errTransactionIDMismatch with no ProtocolRecoveryTimeout set

hsteidl:
This PR fixes a problem where `errTransactionIDMismatch`es are not prevented when no ProtocolRecoveryTimeout is set, even though they are unrelated.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -68,12 +68,13 @@ func (mb *tcpPackager) SetSlave(slaveID byte) {
 }
 
 // Encode adds modbus application protocol header:
-//  Transaction identifier: 2 bytes
-//  Protocol identifier: 2 bytes
-//  Length: 2 bytes
-//  Unit identifier: 1 byte
-//  Function code: 1 byte
-//  Data: n bytes
+//
+//	Transaction identifier: 2 bytes
+//	Protocol identifier: 2 bytes
+//	Length: 2 bytes
+//	Unit identifier: 1 byte
+//	Function code: 1 byte
+//	Data: n bytes
 func (mb *tcpPackager) Encode(pdu *ProtocolDataUnit) (adu []byte, err error) {
 	adu = make([]byte, tcpHeaderSize+1+len(pdu.Data))
 
@@ -100,10 +101,11 @@ func (mb *tcpPackager) Verify(aduRequest []byte, aduResponse []byte) error {
 }
 
 // Decode extracts PDU from TCP frame:
-//  Transaction identifier: 2 bytes
-//  Protocol identifier: 2 bytes
-//  Length: 2 bytes
-//  Unit identifier: 1 byte
+//
+//	Transaction identifier: 2 bytes
+//	Protocol identifier: 2 bytes
+//	Length: 2 bytes
+//	Unit identifier: 1 byte
 func (mb *tcpPackager) Decode(adu []byte) (pdu *ProtocolDataUnit, err error) {
 	// Read length value in the header
 	length := binary.BigEndian.Uint16(adu[4:])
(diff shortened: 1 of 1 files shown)


tcpclient.go
```diff
@@ -68,12 +68,13 @@ func (mb *tcpPackager) SetSlave(slaveID byte) {
...
-//  Unit identifier: 1 byte
-//  Function code: 1 byte
-//  Data: n bytes
+//
+//	Transaction identifier: 2 bytes
+//	Protocol identifier: 2 bytes
+//	Length: 2 bytes
+//	Unit identifier: 1 byte
+//	Function code: 1 byte
+//	Data: n bytes
```
tretmar, line 77: intersting formatting 🤔 I guess this is done automatically by `gofmt`?

tretmar (commented):
Looks good to me. Nice catch! 😇 
What happened to the format? 🤔 

Do we also need to do the same separation in the `default` case?

tcpclient.go
hsteidl: Yes, I did nothing here

hsteidl:
However, as second discussion point but unrelated to this issue: `ProtocolRecoveryTimeout` is a `time.Duration` but never used as such. It's only usage is checking if it's larger than 0. Is this indended? 🤔

andig:
I wrote about `ProtocolRecoveryTimeout` before in https://github.com/grid-x/modbus/pull/60#issue-1437432647. Seems almost as if it could be removed, in any case it's not doing what people think it does...

andig:
Could you also cross-check https://github.com/grid-x/modbus/pull/60 which includes minor optimisation?

gq0 (commented):
LGTM 👍🏻 

As @andig said, @hsteidl could you please cross-check and incorporate the changes from https://github.com/grid-x/modbus/pull/60? ❤️

hsteidl:
I approved and merged #60 and rebased this one onto it. I think this is a solid solution now.
Note that `ProtocolRecoveryTimeout` is still odd but should not be part of these two PRs

tcpclient.go
frzifus: It doesnt look like the `gofmt` default. Looks like your editor replaced 2 spaces in the beginning with one tab. I would recommend to revert it. 

https://tip.golang.org/doc/comment

hsteidl: This occurs because of the two spaces in front of the rows and GoLand refuses to change this behavior.
I even found in https://tip.golang.org/doc/comment at "Syntax" paragraph that `gofmt` causes the reformatting:

> Gofmt indents all lines in a code block by a single tab, replacing any other indentation the non-blank lines have in common.

In our case without bullet points, two spaces belongs to this case and is reformatted.
We could either do
```
// Line1
// Line2
// Line3
```
,
```
// Line1
//   - BulletPoint1
//   - BulletPoint2
```
, or (as in the PR)
```
// Line1
// \tBulletPoint1
// \tBulletPoint2
```
so I recommend to work with one of them, here

hsteidl:
@tretmar Before merging, let's discuss [this](https://github.com/grid-x/modbus/pull/66/files#r1158452083).

tcpclient.go
tretmar: The Bullet Point format fits here IMO

hsteidl:
> @tretmar Before merging, let's discuss [this](https://github.com/grid-x/modbus/pull/66/files#r1158452083).

I created new PR for this: https://github.com/grid-x/modbus/pull/72

tcpclient.go
frzifus: > This occurs because of the two spaces in front of the rows and GoLand refuses to change this behavior.

Definitely something you should fix.

andig: In https://github.com/grid-x/modbus/pull/69/files I've also relied on go-fmt and the result looks fine imho. Maybe we should just go-fmt the entire codebase once plus specifically for the things that were mis-formatted here? Might be easier if this PR was reverted and then re-applied after.

PR #67, opened 2023-04-10
Use embedded fields directly

andig:
This PR aligns and standardises the implementations of ASCII and RTU serial transporters. Before this PR, there was a mix of accessing embedded field's fields directly or via the embedded field. This PR consistently accesses them directly.

It also contains a minor optimisation for variable initialisation in `readIncrementally`.

### DIFF
--- a/asciiclient.go
+++ b/asciiclient.go
@@ -231,7 +231,7 @@ func writeHex(buf *bytes.Buffer, value []byte) (err error) {
 	return
 }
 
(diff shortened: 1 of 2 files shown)


andig:
Ping for review

gq0 (approved):
Thanks for the changes. In general, the coding standard of this library is inconsistent and does not strictly adhere to best practices in Go. Therefore, it requires an overhaul.

PR #81, opened 2024-02-01
Return data of too large responses

### LINKED ISSUE #52: RTU over TCP error: response data size '5' does not match count '4'
I'm trying to read a RTU device routed over TCP using both the library (via InfluxDB Telegraf) and the `modbus-cli` tool. In both cases trying to read the device results in the following error:

```
modbus: response data size '5' does not match count '4'
```

Cli command is:
```
./modbus-cli.amd64 -address tcp://{HOST}:{PORT} -register 40807 -quantity 2 -slaveID 1
```

When switching to RTUoverTCP I get the following timeout error:
```
read tcp 172.24.0.3:52840->{HOST}:{PORT}: i/o timeout
```

Device is SIEMENS SICAM P with Modbus RTU translated over TCP.

Note: When using another tool, for example `modpoll` register value is parsed correctly, e.g.:
```
modpoll -p ${MODBUS_PORT} -a 1 -f -t 4:float -r 40807 ${MODBUS_HOST} 
```
results in:

```
[40807]: 213209248.000000
...
```

Is there some additional parameter I'm missing that needs to be specified to parse correctly the value or is it some issue with the library implementation and the device compatibility?

Thanks!

srebhan:
Some devices send too much data in read responses which is not a grey-zone of  the spec. However, as we do have enough bytes to return, this PR does so as [proposed in the discussion](https://github.com/grid-x/modbus/issues/52#issuecomment-1154381575) . To be able to detect those malformed responses, the PR also introduces a custom error users can check against.

resolves #52

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,15 @@ import (
 	"fmt"
 )
 
+type DataSizeError struct {
+	Expected int
+	Actual   int
+}
+
+func (e *DataSizeError) Error() string {
+	return fmt.Sprintf("modbus: response data size '%d' does not match count '%d'", e.Actual, e.Expected)
+}
+
 // ClientHandler is the interface that groups the Packager and Transporter methods.
 type ClientHandler interface {
 	Packager
@@ -58,7 +67,7 @@ func (mb *client) ReadCoils(address, quantity uint16) (results []byte, err error
 	count := int(response.Data[0])
 	length := len(response.Data) - 1
 	if count != length {
-		err = fmt.Errorf("modbus: response data size '%v' does not match count '%v'", length, count)
+		err = &DataSizeError{Expected: count, Actual: length}
 		return
 	}
 	results = response.Data[1:]
@@ -92,7 +101,7 @@ func (mb *client) ReadDiscreteInputs(address, quantity uint16) (results []byte, 
 	count := int(response.Data[0])
 	length := len(response.Data) - 1
 	if count != length {
-		err = fmt.Errorf("modbus: response data size '%v' does not match count '%v'", length, count)
+		err = &DataSizeError{Expected: count, Actual: length}
 		return
 	}
 	results = response.Data[1:]
@@ -126,7 +135,7 @@ func (mb *client) ReadHoldingRegisters(address, quantity uint16) (results []byte
 	count := int(response.Data[0])
 	length := len(response.Data) - 1
 	if count != length {
-		err = fmt.Errorf("modbus: response data size '%v' does not match count '%v'", length, count)
+		err = &DataSizeError{Expected: count, Actual: length}
 		return
 	}
 	if count != 2*int(quantity) {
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -9,6 +9,17 @@ import (
 	"fmt"
 )
 
+// DataSizeError represents an error for invalid data-sizes i.e. for cases
+// where the data-size does not match the expectation.
+type DataSizeError struct {
+	Expected int
```
andig, line 15: Make clear that his should be in bytes?

srebhan:
@hsteidl anything I can do to get this merged?

srebhan:
@gq0 or @tretmar is there anything I can do to get this PR going forward?

tretmar:
Hello @srebhan , we are currently not able to take the time and assess the changes here thoroughly. It's still not forgotten though ;) We just need time to think about it as this might have direct impact on our product in the end.

srebhan:
Any update here? We need this to fix an actual long-standing issue in Telegraf (https://github.com/influxdata/telegraf/issues/11099) and the effect for existing code is zero if you did proper error checking before...

srebhan:
@hsteidl or @tretmar any update? Another month passed by... This change is **fully backward compatible** and will **NOT** change the current behavior if you did error checking before...

andig:
@srebhan do you want to re-open against https://github.com/evcc-io/modbus? Seems we'll need to go for a permanent fork anyway due to lack of feedback on https://github.com/grid-x/modbus/pull/70.

ValentinMontmirail (commented):
LGTM.

tretmar (approved):
Okay, I understand the idea here now. 

Essentially, you have devices that return more bytes than given within the `Byte count` field of the Modbus Response and you want the library here to be more robust against such cases.

Two things that come into my mind here to improve this even further.
1. Create a function that is doing this lenght/count check for us
 - Currently, there is a lot of duplicate code here. 
 - Not sure, if this is easily doable without some refactoring here.
2. In general, testing of the `client` is missing in this repo completely and it would be really helpful to understand the expectations :D

Still, as this PR doesn't bring any behavioral change other than a different Error type that is being returned, I'm fine with getting this merged first.
I'll then bring the refactoring/test addition into my team. 

As already mentioned quite a lot by @andig , we are not really responsive in this repo and I'm sorry for that. Since the time this repo was started, we've got a lot of new people and never touched this repo at all. Thus, no one feels comfortable or even responsible for this currently.
Sure, this doesn't justify the lack of responsiveness ;)

tretmar:
@srebhan there is a still a conflict that has to be addressed within `client.go`. Could you have a look and ping me afterwards again? 😊

tretmar:
Thanks @srebhan . I'm working on it to get the team more active in here ;)

hsteidl:
Hi @srebhan, we currently don't have a proper release strategy for this repository. I just opened a discussion regarding this, please have a look at #89. 🙂

Issue #50, opened 2022-02-16
Modbus UDP Feature Resquest


theruskian:
## Feature Request

Add the ability to use Modbus/TCP over a UDP transport. 

### Current behavior:
modbus establishes a TCP connection to the modbus slave and polls registers

### Desired behavior:
Send modbus request over UDP to poll registers

### Use case: 
Modbus slaves that support UDP would benefit from the reduced overhead especially when there are multiple masters polling the device for data.

tretmar:
Solved by https://github.com/grid-x/modbus/pull/85

Issue #45, opened 2021-12-25
Seems like this project gets less and less attention and love? Time for some changes?


RTTTC:
It seems like it sometimes takes years to approve pull requests or review issues here@ grid-x/modbus
I do understand that people get busy, life priorities and interests change.
What can be done to give this project a boost and new breath of life, save it from stagnation?
Maybe it would help to have more enthusiastic people that have write privileges and could get things mowing at a faster pace?
Any ideas? What do you guys think?

andig:
@RTTTC I don't feel this is appropiate

RTTTC:
Well , I am new to Github, young and unexperienced.. Too impatient perhaps? So maybe I do not see something? Or do not understand something?
In any case, my questions are not meant to hurt anyone's feelings, I just have a genuine concern for this project and other projects, that depend on this project..

Sadly many projects on Github with time are more and more overlooked or even completely abandoned, and it's natural,  simply such is human nature - as original developers move on to bigger better things in their life, and have no more time or desire to maintain their older projects anymore. 

Maybe there is something that can be done to improve the situation? I think this could be solved simply, as long as there are people that have the knowledge, desire and willingness to join in and help, to keep this going and evolving? 
Or is there no need for changes in your opinion?

gq0:
In terms of responsiveness, we can do better. We will revisit the CI pipeline and PR assignment beginning of next year.

andig:
> In terms of responsiveness, we can do better

@gq0 one PRs open since 2020, a number from 2022. Let me know if I can do anything to help since evcc.io is a popular consumer of this library and we'd be happy to help moving it forward.

andig:
I'm thinking about starting a fork. Would anyone be interested in contributing?

PR #85, opened 2024-02-18
Implement Modbus UDP

StefanNienhuis:
This PR implements Modbus RTU over UDP.

Also implements handling of UDP addresses in Modbus CLI and documents this usage in the README.

This solves #50.

### DIFF
--- a/rtu_over_udp_client.go
+++ b/rtu_over_udp_client.go
@@ -1,0 +1,146 @@
+package modbus
+
+import (
+	"io"
+	"net"
+	"sync"
+)
+
+// RTUOverUDPClientHandler implements Packager and Transporter interface.
+type RTUOverUDPClientHandler struct {
+	rtuPackager
+	rtuUDPTransporter
+}
+
+// NewRTUOverUDPClientHandler allocates and initializes a RTUOverUDPClientHandler.
+func NewRTUOverUDPClientHandler(address string) *RTUOverUDPClientHandler {
+	handler := &RTUOverUDPClientHandler{}
+	handler.Address = address
+	return handler
+}
+
+// RTUOverUDPClient creates RTU over UDP client with default handler and given connect string.
+func RTUOverUDPClient(address string) Client {
+	handler := NewRTUOverUDPClientHandler(address)
+	return NewClient(handler)
+}
+
+// rtuUDPTransporter implements Transporter interface.
+type rtuUDPTransporter struct {
+	// Connect string
+	Address string
+	// Transmission logger
+	Logger logger
+
+	// UDP connection
+	mu   sync.Mutex
+	conn net.Conn
+}
+
+// Send sends data to server and ensures adequate response for request type
+func (mb *rtuUDPTransporter) Send(aduRequest []byte) (aduResponse []byte, err error) {
+	mb.mu.Lock()
+	defer mb.mu.Unlock()
+
+	// Establish a new connection if not connected
+	if err = mb.connect(); err != nil {
+		return
+	}
+
+	// Set write and read timeout
+	// var timeout time.Time
+	// if mb.Timeout > 0 {
+	// 	timeout = mb.lastActivity.Add(mb.Timeout)
+	// }
+	// if err = mb.conn.SetDeadline(timeout); err != nil {
+	// 	return
+	// }
+
+	// Send the request
+	mb.logf("modbus: send % x\n", aduRequest)
+	if _, err = mb.conn.Write(aduRequest); err != nil {
+		return
+	}
+	function := aduRequest[1]
+	functionFail := aduRequest[1] & 0x80
+	bytesToRead := calculateResponseLength(aduRequest)
+
+	var n int
+	var n1 int
+	var data [rtuMaxSize]byte
+	//We first read the minimum length and then read either the full package
+	//or the error package, depending on the error status (byte 2 of the response)
+	n, err = io.ReadAtLeast(mb.conn, data[:], rtuMinSize)
+	if err != nil {
+		return
+	}
+	//if the function is correct
+	if data[1] == function {
+		//we read the rest of the bytes
+		if n < bytesToRead {
+			if bytesToRead > rtuMinSize && bytesToRead <= rtuMaxSize {
+				n1, err = io.ReadFull(mb.conn, data[n:bytesToRead])
+				n += n1
+			}
+		}
+	} else if data[1] == functionFail {
+		//for error we need to read 5 bytes
+		if n < rtuExceptionSize {
+			n1, err = io.ReadFull(mb.conn, data[n:rtuExceptionSize])
+		}
+		n += n1
+	}
+
+	if err != nil {
+		return
+	}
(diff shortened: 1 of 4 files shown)


rtu_over_udp_client.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// Set write and read timeout
+	// var timeout time.Time
+	// if mb.Timeout > 0 {
+	// 	timeout = mb.lastActivity.Add(mb.Timeout)
+	// }
+	// if err = mb.conn.SetDeadline(timeout); err != nil {
+	// 	return
+	// }
```
tretmar, line 57: if not needed, you could even remove it

rtu_over_udp_client.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Send the request
+	mb.logf("modbus: send % x\n", aduRequest)
+	if _, err = mb.conn.Write(aduRequest); err != nil {
+		return
+	}
+	function := aduRequest[1]
```
tretmar, line 64: before accessing `aduRequest[1]`, it's safer to check if the length of `aduRequest` is >= 2.

rtu_over_udp_client.go
```diff
@@ -0,0 +1,146 @@
...
+	functionFail := aduRequest[1] & 0x80
+	bytesToRead := calculateResponseLength(aduRequest)
+
+	var n int
+	var n1 int
+	var data [rtuMaxSize]byte
+	//We first read the minimum length and then read either the full package
```
tretmar, line 71:
```suggestion
	// We first read the minimum length and then read either the full package
```

rtu_over_udp_client.go
```diff
@@ -0,0 +1,146 @@
...
+	//or the error package, depending on the error status (byte 2 of the response)
```
tretmar, line 72:
```suggestion
	// or the error package, depending on the error status (byte 2 of the response)
```

rtu_over_udp_client.go
```diff
@@ -0,0 +1,146 @@
...
+	n, err = io.ReadAtLeast(mb.conn, data[:], rtuMinSize)
+	if err != nil {
+		return
+	}
+	//if the function is correct
+	if data[1] == function {
```
tretmar, line 78: adding a length check for the data slice here would also be safer

tretmar (commented):
I am not deep into the UDP protocol itself and also not into the implementation here in the repo.
As we have no internal use of Modbus RTU via UDP, I also don't see an issue against adding it to the repo here.

I've added a few comments, nothing critical here

rtu_over_udp_client.go
andig: In general- please `gofmt` if not already done ;)

StefanNienhuis:
@tretmar Thanks for the review. I've made the changes.

andig:
What's missing to get this merged?

andig:
@StefanNienhuis after looking through the code (and not being a network protocol guy) I have a situation here in X where we're sending UDP requests to a fixed port and the device replies on a fixed port:

```go
// receiver
addr, err := net.ResolveUDPAddr("udp", "0.0.0.0:8899")
if err != nil {
	return nil, err
}

instance.conn, err = net.ListenUDP("udp", addr)
if err != nil {
	return nil, err
}

// sender
addr, err := net.ResolveUDPAddr("udp", net.JoinHostPort(ip, "8899"))
if err == nil {
        // hand-built RTU
	_, err = m.conn.WriteToUDP([]byte{0xF7, 0x03, 0x89, 0x1C, 0x00, 0x7D, 0x7A, 0xE7}, addr)
}
```

Do you see any chance to support such a scenario? Or maybe I misunderstand this PR or UDP in general?

StefanNienhuis:
@andig I’m not quite sure what your goal is, but I see it’s intended for Goodwe inverters, which is exactly what I built this for. The only thing that would be missing is supporting the AA55 prefix it sends before responses on some models, but that may not be required for your model.

andig:
I think I'm confused by the fact that your code does not have a UDP listener, but I guess I'm just over-complicating UDP mentally.

StefanNienhuis:
Go abstracts away large parts of the UDP protocol details in their network implementation. This allows you to have a UDP ‘connection’ that can send and receive data similar to TCP, even though there isn’t a persistent connection.

tretmar:
Sorry, it is common for us that the authors themselves are merging the PR but I guess you either are not allowed or just not think about merging your own PRs 🤔 

As this has no impact for us, I will merge it now :)

andig:
@StefanNienhuis confirmed working in https://github.com/evcc-io/evcc/discussions/12752#discussioncomment-9127102. Unfortunately we're now confronted with https://github.com/grid-x/modbus/pull/85#issuecomment-2052042978. Seems forking would be the right way to go to implement a workaround for this device defect.

StefanNienhuis:
Yeah, although it's pretty trivial to implement, it may be out of scope for this project to implement such device specific behavior. Not sure who decides what goes in this project and what not, but since you merged this PR, what do you think @tretmar? 

In case you're not familiar, some families of Goodwe inverters accept their commands in normal Modbus format, but send their response in Modbus format with the bytes AA55 prefixed.

tretmar:
I'd say that device specific behavior doesn't belong into this project here 🤔 

Forking brings the risk of getting out of sync really quickly. Another idea is to create an option that you can configure that can deal with the response prefix of AA55. And only use it for your project then.

andig:
> Sorry, it is common for us that the authors themselves are merging the PR but I guess you either are not allowed or just not think about merging your own PRs 🤔

@tretmar nobody contributing here has that permission. See https://github.com/grid-x/modbus/issues/45#issuecomment-1002671524, three years old. There's still a queue of PRs that need attention like the important https://github.com/grid-x/modbus/pull/81.

> Forking brings the risk of getting out of sync really quickly. 

Not responding brings the risk of forking.

> Another idea is to create an option that you can configure that can deal with the response prefix of AA55. And only use it for your project then.

How would I do that without a fork or a merged PR?

tretmar:
> How would I do that without a fork or a merged PR?

Well, either forking or merged PR ofc. Sure, if you face the issue that you have to wait multiple months to get it merged, I understand your concerns.

RTTTC:
> Sorry, it is common for us that the authors themselves are merging the PR but I guess you either are not allowed or just not think about merging your own PRs 🤔

@dammarco - May I ask how have you became a contributor _with rights to merge_ in this project ?
I think it would beneficial for all of us if @andig got similar rights to merge here.
@gq0 - ?

andig:
@dammarco works for gridx. No magic.

> I think it would beneficial for all of us if @andig got similar rights to merge here.

Its perfectly fine for maintainers to do that.

andig:
@StefanNienhuis thank you for adding ModbusUDP. I've noticed that the current implementation does not use any timeouts. This can lead to indefinite wait times, at least in my tests. It would be great, if similar `timeout` to the TCP transporter could be implemented, both for reading and writing.

PR #80, opened 2024-01-05
Enhanced Logging Interface

ValentinMontmirail:
## Overview

This PR introduces an enhancement to the logging system within our codebase. We've evolved from a single-function logging interface to a more robust, level-specific logging interface. This change not only improves the clarity of our logs but also aligns with best practices in logging management.

## Updated Logger Interface: 

The existing logger interface, which previously contained only the Printf function, has been expanded to include three distinct functions: Debug, Info, and Error. Each function is tailored to a specific severity level of logging. The updated interface is as follows:

```go
// logger is the interface to the required logging functions
type logger interface {
    Debug(format string, args ...interface{})
    Info(format string, args ...interface{})
    Error(format string, args ...interface{})
}
```

## Refactoring Log Statements: 

All instances of the Printf method throughout the code have been reviewed and replaced with the appropriate logging level method. This refactoring ensures that each log statement is categorized correctly as per its context and severity. The changes are as follows:

- Debugging logs are now using Debug()
- Informational logs are now using Info()
- Error logs are now using Error()

## Impact:

- Improved Readability: Log statements are now more descriptive and contextually appropriate, making it easier for developers to understand the code flow and debug issues.

- Enhanced Logging Control: With level-specific logging methods, it's now possible to control the verbosity of logs more effectively, which is crucial for production environments.

- Alignment with Logging Standards: This change aligns our logging practices with common industry standards, enhancing the maintainability and scalability of our codebase.

## Request for Review:

I kindly request a review of the changes, focusing on the appropriateness of the log level classifications and the overall structure of the new logging interface. Any feedback or suggestions for further improvement are highly welcome.

## Bug Fixed:

- #79

PR #86, opened 2024-05-02
chore: Bring back ability to use a custom logger

### LINKED ISSUE #79: Enhancement of Logging System in Modbus Package
## Description

The current implementation of the `logger `interface in our Modbus package exclusively uses the `Printf `method for logging. This approach inherently treats all logs as `INFO `level, which is not always appropriate. Many of these logs are more suited to a `DEBUG `level, particularly given their nature and the detail they provide.

## Problem

The primary issue with the current logging system is the lack of granularity. Without distinct log levels (such as `DEBUG`, `INFO`, `ERROR`), it becomes challenging to filter logs based on their severity or importance. This limitation hinders effective debugging and monitoring, especially in production environments where log verbosity needs to be controlled.

## Proposed Solution:

- Introduce a Refined Logger Interface: Implement a more sophisticated logger interface that includes methods for different log levels - Debug, Info, and Error. This change will align our logging practices with standard approaches and provide much-needed flexibility.

- Utilize Standard Logging Libraries: Consider adopting standard logging libraries like Go's slog, which inherently support multiple log levels. This move would not only enhance our logging capabilities but also ensure consistency with common logging practices in the industry.

- Reclassify Existing Logs: Review and reclassify existing log statements in the Modbus package to their appropriate log levels. This reclassification will ensure that logs are more meaningful and contextually relevant.

- Expand Debugging Logs: With the introduction of a Debug log level, we can add more detailed debugging logs throughout the Modbus package. These logs would be invaluable for development and troubleshooting, and they can be easily filtered out in a production setting to reduce noise.

srebhan:
When using this project as a modbus library it is beneficial to be able to use a custom logger to e.g. output the information in a common and consistent form or to use sinks not provided by slog.

This PR abstracts the logging to an interface closely following the notion of `slog.Logger` to allow an easy direct use of slog. As such, the change is **fully backward compatible**. However, this allows users of the library to define custom loggers.

resolves #79

### DIFF
--- a/modbus.go
+++ b/modbus.go
@@ -8,6 +8,7 @@ Package modbus provides a client for MODBUS TCP and RTU/ASCII.
 package modbus
 
 import (
+	"context"
 	"fmt"
 )
 
@@ -115,4 +116,16 @@ type Transporter interface {
 type Connector interface {
 	Connect() error
 	Close() error
+}
+
+// Logger is an interface to define custom loggers
+type Logger interface {
+	Debug(string, ...any)
+	DebugContext(context.Context, string, ...any)
+	Info(string, ...any)
+	InfoContext(context.Context, string, ...any)
+	Warn(string, ...any)
+	WarnContext(context.Context, string, ...any)
+	Error(string, ...any)
+	ErrorContext(context.Context, string, ...any)
 }
--- a/rtu_over_udp_client.go
+++ b/rtu_over_udp_client.go
@@ -3,7 +3,6 @@ package modbus
 import (
 	"fmt"
 	"io"
-	logger "log/slog"
 	"net"
 	"sync"
 )
@@ -46,7 +45,7 @@ type rtuUDPTransporter struct {
 	// Connect string
 	Address string
 	// Transmission logger
-	Logger *logger.Logger
+	Logger Logger
 
 	// UDP connection
 	mu   sync.Mutex
--- a/serial.go
+++ b/serial.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"log/slog"
 	"sync"
 	"time"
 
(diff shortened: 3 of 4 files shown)


andig:
It is really unfortunate that https://github.com/grid-x/modbus/pull/80 was merged _at all_, given its a breaking change. While this PR makes it better, it is still breaking. I'd suggest to revert #80 and instead decide a versioning strategy.
Please don't get me wrong- I'm fine with breaking changes, but:
- #80 has made the logger "interface" much broader than necessary, especially contexts are not used at all in the library and should not be used here
- there are other potentially breaking changes like https://github.com/grid-x/modbus/pull/70 that don't even get any feedback

andig:
The required logger interface for this library would be a single method, not 8. Backwards compatibility is imho not a concern as it was broken with #80 anyway and only very recently.

srebhan:
@andig while I do agree that #80 is unfortunate I also do understand why you want to differentiate between different log-types especially for the CLI. No matter what is decided at the end, the logger should always be an **interface** instead of referencing a fixed implementation to allow the user of the library to choose the logging-framework as you want to have a consistent formatting or log to a file as we do in Telegraf.

andig:
> I also do understand why you want to differentiate between different log-types especially for the CLI

Agreed. But the CLI can do more using an slog or whatever logger without complicating matters for library consumers.

andig:
> Can you outline how you would suppress debug information with just Printf?

Exactly like it was before #80. How you implement the Printf interface is up to the caller. Here‘s what mbmd did and it‘s broken now: https://github.com/volkszaehler/mbmd/blob/master/meters/logger.go

> especially with interface vs func

Either would work. Maybe one day Golang will even auto-implement 1-method interfaces by functions.

srebhan:
@andig I mean how would the modbus library tell you it's a debug message or an error? For a library user this usually is not interesting but for a CLI it might be. Anyway, both ways work for me as long as we can insert our own logging... ;-)

andig:
Afaikt the library only ever sends debug messages?

hsteidl:
Since we just merged #88 which was marked as a replacement for this PR, I'll close it.

Issue #87, opened 2024-05-02
#80 breaks compatibility


andig:
See https://github.com/grid-x/modbus/pull/86 for detailed discussion

PR #88, opened 2024-05-03
Return to minimal logger interface

andig:
Fix https://github.com/grid-x/modbus/issues/87. Replaces https://github.com/grid-x/modbus/pull/86. Refs https://github.com/grid-x/modbus/pull/80#issuecomment-2092654024.

Re-exposes

```go
Printf(format string, v ...any)
```

as minimal logger interface, this time as public type.

Please squash before committing.

/cc @tretmar @srebhan


tretmar (approved):
We (@hsteidl ) looked into it and it looks good to us.

As we are referencing to a specific commit of this repo, we didn't experience any breaking changes.

But if you require this interface and the library anyway only sends debug log messages, then we are fine with doing this ;)

andig:
> Please squash before committing.

@tretmar didn't work. Makes for ugly commit history and potentially harder to merge in forks.

tretmar:
@andig Just to understand you here... you want to have 1 commit per merged PR in the end?

andig:
I think that would be great. At least when the PR does not contain an intentional logical breakdown of the changes.

tretmar:
Okay, we're working internally with a different approach - so that's why we don't squash before merging... simply because we are used to a different approach ;) We follow the approach that the authors of PRs are responsible to have a clean commit structure, including squashing. Still, in a public repository like this here, it might not be the best approach as we can't force everyone to follow this clean commit approach 🤔

srebhan:
@tretmar well you can force people, but why if you can just squash the commits automatically? :-) We are doing this in Telegraf and you can [configure Github](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) to do so.

The big advantage of squashing IMO is that you do have a cleaner commit history keeping things together, but that's a matter of taste of course...

PR #93, opened 2024-09-25
Fix calculating RTU over serial frame delay

andig:
In https://github.com/evcc-io/evcc/issues/16049 we're regularly experiencing timeout errors with multiple devices on a single serial bus. It turns out the frame length calculation according to theMODBUS over Serial Line Specification and Implementation Guide is not correctly implemented. The calculation misses the fact that it needs to honor not only the character gap but also the character duration and it skips the frame interval to the following write.

### DIFF
--- a/rtuclient.go
+++ b/rtuclient.go
@@ -168,7 +168,7 @@ func readIncrementally(slaveID, functionCode byte, r io.Reader, delay time.Durat
 
 	for {
 		if !deadline.IsZero() && time.Now().After(deadline) { // Possible that serialport may spew data
-			return nil, fmt.Errorf("failed to read from serial port within deadline")
+			return nil, errors.New("failed to read from serial port within deadline")
 		}
 
 		if _, err := io.ReadAtLeast(r, buf, 1); err != nil {
--- a/rtuclient.go
+++ b/rtuclient.go
@@ -152,25 +152,33 @@ func (e *InvalidLengthError) Error() string {
 }
 
 // readIncrementally reads incrementally
-func readIncrementally(slaveID, functionCode byte, r io.Reader, deadline time.Time) ([]byte, error) {
+func readIncrementally(slaveID, functionCode byte, r io.Reader, delay time.Duration) ([]byte, error) {
+	if r == nil {
+		return nil, fmt.Errorf("reader is nil")
+	}
+
+	buf := make([]byte, 1)
 	data := make([]byte, rtuMaxSize)
 
 	state := stateSlaveID
 	var length, toRead byte
 	var n, crcCount int
+	var deadline time.Time
 
 	for {
-		if time.Now().After(deadline) { // Possible that serialport may spew data
+		if !deadline.IsZero() && time.Now().After(deadline) { // Possible that serialport may spew data
 			return nil, fmt.Errorf("failed to read from serial port within deadline")
 		}
-		if r == nil {
-			return nil, fmt.Errorf("reader is nil")
-		}
-		buf := make([]byte, 1)
-		_, err := io.ReadAtLeast(r, buf, 1)
-		if err != nil {
+
+		if _, err := io.ReadAtLeast(r, buf, 1); err != nil {
 			return nil, err
 		}
+
+		// after response has started remaining bytes MUST be read within maximum message duration
+		if delay > 0 {
+			deadline = time.Now().Add(delay)
+		}
+
 		switch state {
 		// expecting slaveID
 		case stateSlaveID:
@@ -256,6 +264,11 @@ func (mb *rtuSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, err
 	if err = mb.connect(); err != nil {
 		return
 	}
+
+	// Wait for previous frame delay to elapse
+	time.Sleep(time.Until(mb.lastActivity.Add(2 * mb.frameDelay())))
+	defer func() { mb.lastActivity = time.Now() }()
+
 	// Start the timer to close when idle
 	mb.lastActivity = time.Now()
 	mb.startCloseTimer()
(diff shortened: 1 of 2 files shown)


frzifus (commented):
Does it make sense to may add a small unittest to guarantee we dont break this  behaviour in the future?

rtuclient.go
```diff
@@ -276,19 +285,28 @@ func (mb *rtuSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, err
...
+	var characterDelay int // us
 	if mb.BaudRate <= 0 || mb.BaudRate > 19200 {
 		characterDelay = 750
-		frameDelay = 1750
 	} else {
 		characterDelay = 15000000 / mb.BaudRate
-		frameDelay = 35000000 / mb.BaudRate
 	}
-	return time.Duration(characterDelay*chars+frameDelay) * time.Microsecond
+	return time.Duration((1000000/mb.BaudRate+characterDelay)*chars)*time.Microsecond + mb.frameDelay()
```
andig, line 309:
```suggestion
	return time.Duration((10000000/mb.BaudRate+characterDelay)*chars)*time.Microsecond + mb.frameDelay()
```

andig:
> Does it make sense to may add a small unittest to guarantee we dont break this behaviour in the future?

Updated the comment. Instead of the freaky constants the "proper" fix would be something like https://github.com/evcc-io/modbus/pull/9 which calculates the actual number of bits and hence delay.

For time being I've tried to keep it simple and non-invasive.

frzifus (approved):

led0nk (approved):

andig:
Thanks for the review. Code and test updated once more for proper 11bit encoding instead of the assumed 8N1 which is illegal in terms of modbus spec (https://stackoverflow.com/questions/20740012/calculating-modbus-rtu-3-5-character-time). 

Good to merge from evcc point of view. Thank you!

rtuclient.go
```diff
@@ -276,19 +285,29 @@ func (mb *rtuSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, err
...
+	return time.Duration(frameDelay) * time.Microsecond
+}
+
+// calculateDelay roughly calculates time needed for the next frame.
+func (mb *rtuSerialTransporter) calculateDelay(chars int) time.Duration {
+	var characterDelay int // us
+	if mb.BaudRate <= 0 || mb.BaudRate > 19200 {
+		characterDelay = 750
+	} else {
+		characterDelay = 16500000 / mb.BaudRate
```
premultiply, line 308:
```suggestion
		characterDelay = 11000000 / mb.BaudRate
```

The factor 1.5 is wrong here.
T1.5 is the gap timeout on the character stream.
That must be handled inside the stream parser.

rtuclient.go
```diff
@@ -276,19 +285,29 @@ func (mb *rtuSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, err
...
 	}
-	return time.Duration(characterDelay*chars+frameDelay) * time.Microsecond
+	return time.Duration((11000000/mb.BaudRate+characterDelay)*chars)*time.Microsecond + mb.frameDelay()
```
premultiply, line 310:
```suggestion
	return time.Duration(characterDelay*chars)*time.Microsecond + mb.frameDelay()
```
Only the minimum transmission time of each character is relevant here.
The whole calculation only substracts transmission time from time slaves processing delay.
The processing time of the slave is not known.
The maximum processing timeout is `[timeout]+[calculated minimum transmission time].`

rtuclient.go
```diff
@@ -276,19 +285,29 @@ func (mb *rtuSerialTransporter) Send(aduRequest []byte) (aduResponse []byte, err
 	return
```
premultiply, line 285:
```suggestion
	time.Sleep(mb.frameDelay())
	return
```
is missing here, as the parser returns just after the _calculated_ last character is received instead of waiting for >=T3.5 gap in datastream.
Ref https://www.modbus.org/docs/PI_MBUS_300.pdf Page 9

**This** finally fixes the read errors with multiple devices on the bus without fixing the parser itself.

andig:
Point of the current implementation is to wait the maximum duration of the response message. It is not a full RTU parser. What it (still) does not do correctly within that limitation is to wait for client response start which must be _at least_ 3.5 characters after the request. This is likely too short if the slave is slow to respond. I‘ll look into adding that as well, keeping the diff as low as possible.

rtuclient.go
andig: Waiting for the gap after receive happens before the next write. See above.

andig:
PR has been updated to take slave response delay into account. Before replying, the slave may take any time to prepare it's response. Waiting for a fixed amount of time (send data duration, gap, receive data duration) is only sufficient if client responds more or less immediately after the gap. Instead, with last commit the receive data duration is only considered after first byte has been is received. Since `delay` includes that first byte and the following frame delay that should provide headroom for not stopping read too early.

RTTTC:
So, is this improvement finished  completely and is ready to merge ?

andig:
@premultiply and me have revised and thoroughly tested the changes. We're confident that this implementation is working. One tiny bit that remains unclear is why we need 2*framedelay for the gap, but with lower gap we could not get stable transmission with multiple devices on the bus.

Would be great if this could get merged.

andig:
Includes https://github.com/grid-x/modbus/pull/94

rtuclient.go
```diff
@@ -164,10 +164,11 @@ func readIncrementally(slaveID, functionCode byte, r io.Reader, deadline time.Ti
 	state := stateSlaveID
 	var length, toRead byte
 	var n, crcCount int
+	var deadline time.Time
 
 	for {
-		if time.Now().After(deadline) { // Possible that serialport may spew data
-			return nil, errors.New("failed to read from serial port within deadline")
+		if !deadline.IsZero() && time.Now().After(deadline) { // Possible that serialport may spew data
+			return nil, fmt.Errorf("failed to read from serial port within deadline")
```
andig, line 171:
```suggestion
			return nil, errors.New("failed to read from serial port within deadline")
```

andig:
Friendly ping. PR is ready from hour side and already merged in the evcc fork.

RTTTC:
Maybe @dammarco could have a look ?

dammarco:
Hey first of all, sorry for the long silence...
I already brought it to the team when you first raised it, but there was no time yet to review it.
I'll try to look into it today 😊

RTTTC:
??

dammarco:
Hey @andig , why was it closed now? 🤔 Not necessary anymore?

andig:
A bit of an explanation: this repo does not ensure RTU serial timings. It disguises that by calculating character durations (https://github.com/grid-x/modbus/blob/master/rtuclient.go#L274) but in reality just waits for serial timeout (https://github.com/grid-x/modbus/blob/master/rtuclient.go#L276). It also does not ensure frame delays which makes it unsuitable for parallelising bus operations. This can easily be fixed by https://github.com/evcc-io/modbus/blob/master/rtuclient.go#L265-L267. Fixing the former- RTU serial timings- we've corrected the calculation (11 bits, missing character duration etc). However, we've found that some popular devices violate the timings according to the ModBus implementation guide (like Heidelberg Energy Control). Since we need to support these we've abandoned the PR and went back to the original non-compliant design but added proper frame delays.

PR #96, opened 2025-03-12
Prevent falsy comparison against nil `net.Conn`

antoineco:
When the result of `net.Dial()` is assigned to `mb.conn` unconditionally, `mb.conn` will have the value `net.Conn(nil)` instead of an explicit nil value[^1] when the function fails (returns an error). As a result, comparisons to nil are falsy and panics may occur due to nil pointer dereferences.

[^1]: https://100go.co/#returning-a-nil-receiver-45

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -350,20 +350,19 @@ func (mb *tcpTransporter) Connect() error {
 
 func (mb *tcpTransporter) connect() error {
 	if mb.conn == nil {
-		dialer := net.Dialer{Timeout: mb.Timeout}
+		var conn net.Conn
 		var err error
+		dialer := net.Dialer{Timeout: mb.Timeout}
 		if mb.tlsConfig != nil {
-			mb.conn, err = tls.DialWithDialer(&dialer, "tcp", mb.Address, mb.tlsConfig)
-			if err != nil {
+			if conn, err = tls.DialWithDialer(&dialer, "tcp", mb.Address, mb.tlsConfig); err != nil {
 				return err
 			}
 		} else {
-			mb.conn, err = dialer.Dial("tcp", mb.Address)
-			if err != nil {
+			if conn, err = dialer.Dial("tcp", mb.Address); err != nil {
 				return err
 			}
 		}
-
+		mb.conn = conn
 		// silent period
 		time.Sleep(mb.ConnectDelay)
 	}


andig:
I'm confused as to what problem this PR tries to solve. As long as `net` returns proper nil (and not a differently dynamically typed nil- which we can rely on) the old and simpler code is perfectly valid. See https://play.golang.com/p/kYoFYdqok0O.

antoineco:
@andig you're making a bold assumption, the dialer is not guaranteed to return a proper `nil`.

Here is the counter example to your snippet: https://go.dev/play/p/gFy1EQ2DMBV

andig:
Thanks for pointing this out. Quickly stepped through in the debugger:

```go
	rawConn, err := netDialer.DialContext(ctx, network, addr)
	if err != nil {
		return nil, err
	}
```

This is where the result conn is coming from. I'm still puzzled how this applies to go100...

> When returning an interface, be cautious about not returning a nil pointer but an explicit nil value

...because this _is_ an explicit nil value.

antoineco:
If I'm not mistaken the problem is on our end. If we assign the way we did before this commit (without a nil check), `m.conn` is no longer a true `nil`. It's a _very_ misleading Go behavior, and not the first time I encounter it.

antoineco:
The debugger actually provides a good explanation here.

Before assigning to `conn`, the value is `nil`:

```gdb
(dlv) n
> main.main() ./main.go:12 (PC: 0x101150e7c)
     7: )
     8:
     9: var conn net.Conn
    10:
    11: func main() {
=>  12:         conn, _ = tls.Dial("invalid", "invalid", &tls.Config{})
    13:         if conn == nil {
    14:                 fmt.Println("nil")
    15:         }
    16:
    17: }
```

```gdb
(dlv) p conn
net.Conn nil
```

After assigning to `conn`, the value is an interface with a `nil` value instead:

```gdb
(dlv) n
> main.main() ./main.go:13 (PC: 0x101150f70)
     8:
     9: var conn net.Conn
    10:
    11: func main() {
    12:         conn, _ = tls.Dial("invalid", "invalid", &tls.Config{})
=>  13:         if conn == nil {
    14:                 fmt.Println("nil")
    15:         }
    16:
    17: }
```

```gdb
(dlv) p conn
net.Conn(*crypto/tls.Conn) nil
```

The Go100 explanation is a bit vague, but it is indeed the same problem we encountered here.

andig:
I guess this play shows what's going on and what go100 means by "not returning nil pointer": https://go.dev/play/p/WGwvQhDtuBe. tls returns a nil pointer, not an interface. That changes the `net.Conn` interface variable's dynamic type to `*tls.Conn` which in turn makes it non-nil.

antoineco:
I was made aware that the official Go FAQ has a great description about this behaviour as well: https://go.dev/doc/faq#nil_error. IMO a much better one than Go 100 :)

PR #98, opened 2025-03-26
chore: revert reordering

andig:


antoineco:
This seems to have been intentional, or at least automatically done by the language server.

The default rule typically orders packages as follows:

```
<stdlib>

<third party>

<module>
```

And `github.com/grid-x/modbus` turns out to be this module:

https://github.com/grid-x/modbus/blob/d1d8b421f52b16001ca1e289746ef7b998b13214/go.mod#L1

andig:
Afaia this is not done by language server but by gofmt. And gofmt does- to my knowledge- not have a rule to separate the module. If the module (really any module) is separated into a new block, gofmt will respect this though. Removing the empty line and re-running gofmt should confirm that.

antoineco:
This change wasn't done by me, but I tried to reproduce it by putting the file in the following state:

```patch
diff --git i/cmd/modbus-cli/main.go w/cmd/modbus-cli/main.go
index 1399c59..f7dbb0f 100644
--- i/cmd/modbus-cli/main.go
+++ w/cmd/modbus-cli/main.go
@@ -15,10 +15,6 @@ import (
        "strings"
        "text/tabwriter"
        "time"
-
-       "github.com/grid-x/serial"
-
-       "github.com/grid-x/modbus"
 )

 func main() {
```

And after sending a `source.organizeImports` code action to `gopls`, it modified the file in the same way as it is right now (default `gopls` config).

I would suggest that we don't perform the change you suggested, otherwise it will be reverted by the next editor.

andig:
When I remove the imports and save the file in VScode it will be as PR'ed. I can only say that I'm not aware of intentional changed module ordering in either gopls or gofmt.

antoineco:
I found the relevant setting: https://github.com/golang/tools/blob/gopls/v0.18.1/gopls/doc/settings.md#local-string

You're right, it was set by my editor and not by `gopls` directly: `["local"] = "github.com/grid-x/modbus"`.
[GoLand](https://www.jetbrains.com/go) defaults to that as well.

We should have better CI checks in place to enfore this kind of formatting/import rules and avoid frictions.

PR #99, opened 2025-04-02
Allow overriding Dial function in TCP client handler

antoineco:
Allows customizing the Client function that is used for dialing new connections.

The primary envisioned use case is to enable the outsourcing of network connections to an external connection manager (pool):

```go
var cm extpkg.ConnectionManager

modbus.NewTCPClientHandler(
    modbus.WithDialer(func(ctx context.Context, _, addr string) (net.Conn, error) {
        // acquire new connection inside the closure
       return cm.TCPConn(ctx, address)
    })
)
```

The default Dial function remains `net.Dial`.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -5,6 +5,7 @@
 package modbus
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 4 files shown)


tcpclient.go
```diff
@@ -49,12 +50,26 @@ func NewTCPClientHandler(address string, options ...TCPClientHandlerOption) *TCP
 	h.Address = address
 	h.Timeout = tcpTimeout
 	h.IdleTimeout = tcpIdleTimeout
+	if h.Dial == nil {
+		h.Dial = (&net.Dialer{Timeout: h.Timeout}).DialContext
```
mochaibi, line 54: nit: the instantiation of `net.Dialer{Timeout: timeout}.DialContext ` is repeated across multiple handlers. It can be extracted into a common function, which can be reused by all handlers

mochaibi (approved):

joschahenningsen (approved):

tcpclient.go
antoineco: Just to make sure I understand your comment, are you suggesting something like this?
```go
func defaultDialFunc(timeout time.Duration) DiaFunc {
    return (&net.Dialer{Timeout: timeout}).DialContext
}

func NewTCPClientHandler(/* ... */) {
    // ...
    if h.Dial == nil {
        h.Dial = defaultDialFunc(h.Timeout)
    }
}
```

mochaibi: Yess 🙂

antoineco: Done in [c7750a0..29e33de](https://github.com/grid-x/modbus/compare/c7750a0aaa1cc1df9d6b60420580557488b3a752..29e33de831776479f300cff8de97e6131e47f432)

PR #101, opened 2025-04-03
Allow disabling connection caching in TCP client handler

antoineco:
When `IdleTimeout` is zero, only use dialed connections for a single Send.
This is useful when paired with a custom `Dial` function (#99) to ensure that the client returns the connection immediately instead of trying to reuse it.

When `IdleTimeout` is _negative_, we keep treating it as a disable switch (connections don't expire). This logic remains unchanged.

<details><summary>Outdated comment</summary>
<p>

_My initial approach introduced a new parameter, `DisableConnCaching`, which I wasn't particularly fond of._

---

I would have very much preferred to use `IdleTimeout ≦ 0` for this, but this would be a breaking change because `IdleTimeout ≦ 0` currently means "cache forever" (2 usages):

https://github.com/grid-x/modbus/blob/2deb806569592e3da22fe4c8b08f96515c7d85a9/tcpclient.go#L390-L399

https://github.com/grid-x/modbus/blob/2deb806569592e3da22fe4c8b08f96515c7d85a9/tcpclient.go#L440-L453

</p>
</details>

### DIFF
--- a/tcpclient_test.go
+++ b/tcpclient_test.go
@@ -13,6 +13,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"math/big"
@@ -191,7 +192,7 @@ func TestCustomDialer(t *testing.T) {
 	const tSentinelVal uint32 = 0xBADC0DE
 	const qtyUint32 = 2
 
-	// Processes cli.ReadInputRegisters() and returns a static integer value.
+	// Processes a single cli.ReadInputRegisters() and returns a static integer value.
 	acceptConnAndRespond := func(srvLn net.Listener) error {
 		conn, err := srvLn.Accept()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


andig:
Why no BC? Not a stable version?

andig:
Not sure. Seems a bit unexpected?

antoineco:
I actually just amended the PR with the approach I suggested in my previous comment. The more I think about it, the more it makes sense to me.

- `IdleTimeout` is a non-optional attribute and the default constructor for `TCPClientHandler` sets it to a positive value (60 seconds).
- If a caller switches `IdleTimeout` to `0` explicitly, it seems acceptable to expect the timeout to have an _immediate_ effect.
- Conversely, negative values are often encoutered to disable a behavior when the setting isn't expected to have a negative value (under normal conditions). Timeout durations aren't expected to be negative, so I think we can keep treating it as an "off" value.

IMO this approach is still more legible than introducing yet another parameter, and easily clarified with good documentation. It is also implemented with very minor changes to the existing code, which carries a lot of legacy.

tcpclient_test.go
```diff
@@ -348,6 +349,174 @@ func TestCustomDialer(t *testing.T) {
 	})
 }
 
+func TestConnCaching(t *testing.T) {
```
joschahenningsen, line 352: Very extensive, nice

joschahenningsen (approved):
Approach is fine with me 👍

andig:
@antoineco I've had another look at the wording and I find it quite confusing.

> If negative, cached connections do not time out.

What does "cached connections" refer to here? There is only a single connection, what makes this "cached"? Do you mean "connection is not closed when idle"?

> If zero, disables the caching of TCP connections and only uses dialed connections for a single Send.

What does "caching" refer to here? Do you mean that a closed connection won't be dialed again and all further requests will result in error? What means "a single send" in that aspect?

Imho- if idle timeout has been disabled- this would typically have been done by setting it to 0. Changing the meaning of the zero timeout value is quite unfortunate.

antoineco:
_Caching_ here refers to the fact of caching a connection for future re-use.

I deliberately used the same wording as in the [net/http](https://pkg.go.dev/net/http#Transport) package, because the feature this PR enables is similar to what the net/http Transport enables with regards to connection caching (with the difference that the Modbus client opens between 0 and 1 connection, not 0:n).

PR #103, opened 2025-05-16
fix: Update .goreleaser.yaml to v2

joschahenningsen:
Bumps the version of `.goreleaser.yaml` to v2 and fixes deprecations:

- DEPRECATED:  snapshot.name_template  should not be used anymore, check https://goreleaser.com/deprecations#snapshotname_template for more info
- DEPRECATED:  archives.format_overrides.format  should not be used anymore, check https://goreleaser.com/deprecations#archivesformat_overridesformat for more info

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,3 +1,4 @@
+version: 2
 builds:
   - main: ./cmd/modbus-cli
     env:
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -26,11 +27,11 @@ archives:
       {{- else }}{{ .Os }}{{ end }}_{{ .Arch }}
     format_overrides:
     - goos: windows
-      format: zip
+      formats: [ 'zip' ]
```
antoineco, line 30: nit:
```suggestion
      formats: [ zip ]
```
Quotes are optional around strings

antoineco (approved):

.goreleaser.yaml
joschahenningsen: Right, I just went for the suggestion from the deprecation guide here.

Issue #97, opened 2025-03-24
TCP: better handling of server-closed connections


andig:
In https://github.com/evcc-io/evcc/discussions/12852#discussioncomment-12606558, we're having problems with some modbus devices that have issues on ModbusTCP:

- device is resource-limited
- it quickly closes idle connections (device timeout 4s)
- library will only notice the closed connection on the next send

What we notice it that it seems to take very long before the transaction is retried:

```
[grid1 ] TRACE 2025/03/24 16:59:06 modbus: send 00 0f 00 00 00 06 ff 03 04 36 00 01
[grid1 ] TRACE 2025/03/24 16:59:06 modbus: close connection and retry, because of EOF
[grid1 ] TRACE 2025/03/24 16:59:21 modbus: send 00 0f 00 00 00 06 ff 03 04 36 00 01
```

Does anyone have similar issues or experience with ModbusTCP against low-spec embedded devices? Are the best practices for (timeout) settings to improve handling dropped connections?

andig:
I've added a bit more logging and found that, when the server closes the connection, then:

- the send on the (already closed) connection seems to succeed (not sure why this happens- TCP FIN/ACK/RST magic?)
- the response read fails in `readResponse` with `EOF`
- this causes `readResponse` to return `readResultCloseRetry`, which leads to `time.Sleep(mb.LinkRecoveryTimeout)`

Imho, in the case of `EOF`, `readResponse` should return `readResultRetry` and connection being immediately retried.

As workaround, I've solved this by setting `mb.LinkRecoveryTimeout` to `0`.

andig:
To be more specific- since `readResultRetry` applies to retrying without closing the connection- we should probably do something like:

```go
mb.logf("modbus: close connection and retry, because of %v", err)

mb.close()
if err != io.EOF {
    // make timeout conditional
    time.Sleep(mb.LinkRecoveryTimeout)
}
```

andig:
I've just hin another instance- the infamous `connection reset by peer`/`broken pipe`: https://gosamples.dev/connection-reset-by-peer/.

When this happens, reading the response will fail in `readResponse`:

```go
if _, err = io.ReadFull(mb.conn, data[:tcpHeaderSize]); err == nil {
        ...
} else if (err != io.EOF && err != io.ErrUnexpectedEOF) ||
	mb.LinkRecoveryTimeout == 0 || time.Until(recoveryDeadline) < 0 {
	return
}
```

[...]

PR #104, opened 2025-07-01
Handle broken pipe error in TCP client

alexjoedt:
Close the connection when a broken pipe error occurs to ensure a new connection can be established on the next attempt.

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -8,11 +8,13 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"net"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 )
 
@@ -231,6 +233,12 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
 		// Send data
 		mb.logf("modbus: send % x", aduRequest)
 		if _, err = mb.conn.Write(aduRequest); err != nil {
+			if errors.Is(err, syscall.EPIPE) {
+				// If a "broken pipe" (EPIPE) error occurs, it means the remote side has closed the connection.
+				// In this case, we close our side of the connection as well, so that the next attempt will establish a new connection,
+				// because [mb.connect] only dials if [mb.conn] is nil
+				mb.close()
+			}
 			return
 		}
 


alexjoedt:
The fix is intended to ensure that `mb.connect` actually has an effect when an error occurs.
However, I think there is a deeper issue with the current logic.
If the peer closes the connection, `mb.conn` is never set to nil, so calling [mb.connect](https://github.com/grid-x/modbus/blob/4b99c910e8e77046bf6ed08386c09847a21f2fb4/tcpclient.go#L386) has no effect.

Is this behavior intentional?

If not we might need another specific error handling while read from the connection.

@philippfranke in case of remote errors your [retry logic ](https://github.com/grid-x/impact/blob/c281bf37f34da91deace8a6d8a27c03836d37168/protocol/retry.go#L34)will have no effect, right?

@antoineco Since you're deep in the subject here because of parallel scanning, what do you think?

antoineco:
This was raised in #97 and yes, the current error handling (including retries) is suboptimal, even wrong depending on the timeout values selected.

I'm with @andig here. Instead of trying to be smart by stacking error conditions we should probably just simplify things as much as possible and let the client retry on most network errors.

@andig how has this worked out so far in EVCC? I reckon that your simplification of that code was a success?

---

On a different topic, I'm surprised to see `EPIPE` in the wild. Usually `EOF` is returned by Go whenever the peer closes the connection.

alexjoedt:
> On a different topic, I'm surprised to see `EPIPE` in the wild. Usually `EOF` is returned by Go whenever the peer closes the connection.

Yes,`EOF` is raised when you read from a connection that was closed from the peer. But when you try to write to a connection that is closed, you will get a broken pipe error.

---

> This was raised in #97 and yes, the current error handling (including retries) is suboptimal, even wrong depending on the timeout values selected.
> 
> I'm with @andig here. Instead of trying to be smart by stacking error conditions we should probably just simplify things as much as possible and let the client retry on most network errors.
> 
> @andig how has this worked out so far in EVCC? I reckon that your simplification of that code was a success?
> 

Oh i see, you targeting exactly this problem in #97.

But i am wondering, that every call to the handlers connect has no effect when the server resets the connection.
I think we should tackle this asap, because i am afraid that we might have an issue in one of our EVCS driver that is related to this issue.

antoineco (commented):
One comment, otherwise LGTM since it addresses a production bug.

We should still tackle the shortcomings of the TCP error handling, but can do that in a separate PR.

andig:
Will provide feedback later on, delayed due to vacation.

alexjoedt:
> One comment, otherwise LGTM since it addresses a production bug.
> 
> We should still tackle the shortcomings of the TCP error handling, but can do that in a separate PR.

Yes, I agree we should do this.

However, as mentioned, I think we need separate handling for when the connection read returns io.EOF.

tcpclient.go
```diff
@@ -231,6 +233,14 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
 		// Send data
 		mb.logf("modbus: send % x", aduRequest)
 		if _, err = mb.conn.Write(aduRequest); err != nil {
+			if errors.Is(err, syscall.EPIPE) {
+				// If a "broken pipe" (EPIPE) error occurs, it means the remote side has closed the connection.
+				// In this case, we close our side of the connection as well, so that the next attempt will establish a new connection,
+				// because [mb.connect] only dials if [mb.conn] is nil
+				mb.mu.Lock()
```
andig, line 240: The locking seems odd. Below in https://github.com/grid-x/modbus/pull/104/files#diff-49830af181541bc76c8bca9e4401d808240a559716536397ae4d22947acddf32R262 we don't lock?

alexjoedt:
> @alexjoedt has this been tested against an affected appliance?

Nope, the connection to the affected asset is blocking new connections for a while after an invalid connection attempt (OEM will fix this).

I still think this is a good fix, because if the server closes the connection, we can establish a new one the next time we try.

PR #110, opened 2025-10-29
feat: implement Read Device Identification specific object

jhorck:


antoineco:
@jhorck thanks!

Out of curiosity, what object ID are you trying to read that is not part of the cases supported by `ReadDeviceIdentification()`?

antoineco (approved):

Issue #114, opened 2025-11-17
TCP connection error handling logic


sansmoraxz:
I am trying to understand the link recovery mechanism. So for this, I am using a tcp modbus simulator. Sometimes I restart the server, or disconnect/ reconnect the simulator into the network.

So here's what I have observed. io.EOF is being propagated into the clients.

So, is connection recovery a built in feature of the library, or should be handle those from client side (by dropping the previous client).

Going deeper into the codebase, I see this:

https://github.com/grid-x/modbus/blob/99e372e638c1fce69f9ab43df4531c16deba4a74/tcpclient.go#L321-L328

The condition `(err != io.EOF && err != io.ErrUnexpectedEOF)` check feels convoluted. Also, `readResultCloseRetry` is not used anywhere else other than this assignment.

Feels like a logic bug, but I am not sure of the intention here too.

andig:
This fits my observations (can‘t find the comment atm). Link recovery doesn‘t do what it is supposedly designed to do.

PR #115, opened 2025-11-17
Fix and refactor TCP connection handling

sansmoraxz:
- Improve IO errors handling
- Use proper timeout values for io issues
- Some improved logging

Fixes: #114


sansmoraxz:
> Would this potentially also solve https://github.com/grid-x/modbus/issues/97?

I expect it to.

I just tested by restarting diagslave instances. And keeping persistent connection.

PR #118, opened 2025-11-20
Revert "TCP improve protocol deadline checks"

mkrafeld:
Reverts grid-x/modbus#116

We released this feature to production and noticed a panic.

From our analysis: 
`processResponse` can return the error `ErrTCPHeaderLength` and this won't be caught in the if 
`if err == io.EOF || err == io.ErrUnexpectedEOF || err == syscall.ECONNRESET { ... }` 
leading to call `err = verify(aduRequest, aduResponse)` with an empty `aduResponse` byte array. 
This then calls `responseVal := binary.BigEndian.Uint16(aduResponse)` which leads to the panic. 

Stack trace:
```
2025-11-20 11:48:56.037 github.com/grid-x/modbus.(*tcpTransporter).readResponse(0x38be408, {0x35e6ef0, 0xc, 0xc}, {0x346fe60, 0x104, 0x104}, {0xc23fda65c1c048e3, 0xc3b544bfe, 0x1b2fa60}, ...)
2025-11-20 11:48:56.037	/app/vendor/github.com/grid-x/modbus/tcpclient.go:373 +0x264
2025-11-20 11:48:56.037	github.com/grid-x/modbus.verify({0x35e6ef0, 0xc, 0xc}, {0x0, 0x0, 0x0})
2025-11-20 11:48:56.037	/usr/local/go/src/encoding/binary/binary.go:156
2025-11-20 11:48:56.036	encoding/binary.bigEndian.Uint16(...)
2025-11-20 11:48:56.036	goroutine 883 [running]:
2025-11-20 11:48:56.034 panic: runtime error: index out of range [1] with length 0
```

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -214,7 +214,6 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
 	linkRecoveryDeadline := time.Now().Add(mb.LinkRecoveryTimeout)
 	protocolRecoveryDeadline := time.Now().Add(mb.ProtocolRecoveryTimeout)
 
-	isPrevReadErr := false
 	for {
 		// Establish a new connection if not connected
 		if err = mb.connect(ctx); err != nil {
@@ -258,19 +257,8 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
 			return
 		case readResultRetry:
 			mb.logf("modbus: retry reading response, because of %v", err)
-			if isPrevReadErr {
-				// two read errors in a row - close connection to force with fresh transactions
-				mb.close()
-			}
-			isPrevReadErr = true
-			select {
-			case <-ctx.Done():
-				return nil, ctx.Err()
-			default:
-				continue
-			}
+			continue
 		case readResultCloseRetry:
-			isPrevReadErr = false
 			mb.logf("modbus: close connection and retry reading response, because of %v", err)
 			mb.close()
 			select {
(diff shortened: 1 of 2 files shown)


antoineco (approved):

dammarco (approved):

sansmoraxz:
lemme check

sansmoraxz:
> From our analysis: `processResponse` can return the error `ErrTCPHeaderLength` and this won't be caught in the if `if err == io.EOF || err == io.ErrUnexpectedEOF || err == syscall.ECONNRESET { ... }` leading to call `err = verify(aduRequest, aduResponse)` with an empty `aduResponse` byte array. This then calls `responseVal := binary.BigEndian.Uint16(aduResponse)` which leads to the panic.
> 
> Stack trace:
> 
> ```
> 2025-11-20 11:48:56.037 github.com/grid-x/modbus.(*tcpTransporter).readResponse(0x38be408, {0x35e6ef0, 0xc, 0xc}, {0x346fe60, 0x104, 0x104}, {0xc23fda65c1c048e3, 0xc3b544bfe, 0x1b2fa60}, ...)
> 2025-11-20 11:48:56.037	/app/vendor/github.com/grid-x/modbus/tcpclient.go:373 +0x264
> 2025-11-20 11:48:56.037	github.com/grid-x/modbus.verify({0x35e6ef0, 0xc, 0xc}, {0x0, 0x0, 0x0})
> 2025-11-20 11:48:56.037	/usr/local/go/src/encoding/binary/binary.go:156
> 2025-11-20 11:48:56.036	encoding/binary.bigEndian.Uint16(...)
> 2025-11-20 11:48:56.036	goroutine 883 [running]:
> 2025-11-20 11:48:56.034 panic: runtime error: index out of range [1] with length 0
> ```



On happy path there are guards:
```go
		aduResponse, err = mb.processResponse(data[:])
		if err == nil {
			err = verify(aduRequest, aduResponse)
			if err == nil {
				mb.logf("modbus: recv % x\n", aduResponse)
				return // everything is OK
			}
		}
```

mkrafeld:
> 
> On happy path there are guards:
> 
> ```go
> 		aduResponse, err = mb.processResponse(data[:])
> 		if err == nil {
> 			err = verify(aduRequest, aduResponse)
> 			if err == nil {
> 				mb.logf("modbus: recv % x\n", aduResponse)
> 				return // everything is OK
> 			}
> 		}
> ```

Yeah but those guards were exchanged with specific error handling. The `ErrTCPHeaderLength` was not part of that guard anymore with led to the panic at a later place 😄

sansmoraxz:
I suspect something it's something in the `verify` method triggered it. Cause it would have never got into that block itself. `processResponse`'s err value is `nil` in your case.

But anyway lemee make it cleaner.

sansmoraxz:
Also is there any way to reproduce the panic into a test case? That would help out a lot in debugging this.

mkrafeld:
> Also is there any way to reproduce the panic into a test case? That would help out a lot in debugging this.

Sadly we did not yet got to the point to reproduce this issue. We just saw panics at this place in the code 😅

sansmoraxz:
What I think happened is you got err as nil and the aduResponse as empty slice (somehow) from `processResponse`

mkrafeld:
> What I think happened is you got err as nil and the aduResponse as empty slice (somehow) from `processResponse`

This is also was we think from our first investigation. But can't it also happen that `processResponse` returns `ErrTCPHeaderLength` as an error? In that case we would not catch it in the if (before this revert) and would have an empty `aduResponse ` 😄

sansmoraxz:
I mean that's what the first guard is in nested condition. `if err == nil {` it would never get into that block to execute verify if some error value was returned. It would just have skipped ahead to the deadline check and if it checks out over to the error type check.

sansmoraxz:
IMO the function needs another refactor. The error checks should ideally be right next to their caller.

mkrafeld:
But looking at the previous code it was as follows:
```
// err could be ErrTCPHeaderLength, right? That would not be caught by the if below
aduResponse, err = mb.processResponse(data[:])
if err == io.EOF || err == io.ErrUnexpectedEOF || err == syscall.ECONNRESET {
	mb.logf("modbus: connection closed by remote side: %v", err)
	// recovery disabled or deadline reached - report error
	if mb.LinkRecoveryTimeout == 0 || time.Until(recoveryDeadline) < 0 {
		return
	}
	res = readResultCloseRetry
	return

}

// calling verify with an aduResponse which might be empty as we do not check every possible error
err = verify(aduRequest, aduResponse)
```

I might be missing something though if this error path is not reachable at all 😄

sansmoraxz:
Also some functions are returning via fmt.Errorf which would skip over the respective type check

sansmoraxz:
I just though of something. Is there a guard check against `aduRequest`?

sansmoraxz:
What if empty bytes passed into it?

mkrafeld:
I'm not sure. I just stumbled across this panic and therefore reverted the PR 😅

PR #120, opened 2025-11-20
Fix: more guards and refactor tcp readResponse 

sansmoraxz:
Improvements on #118

### DIFF
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -288,54 +288,50 @@ func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryD
 			}
 			return
 		}
-		aduResponse, err = mb.processResponse(data[:])
-		if err == nil {
-			err = verify(aduRequest, aduResponse)
-			if err == nil {
-				mb.logf("modbus: recv % x\n", aduResponse)
-				return // everything is OK
+		aduResponse, err = mb.processResponse(data[:]) // this also does io
+		if err != nil {
+			// recovery disabled or deadline reached - report error
+			if mb.LinkRecoveryTimeout == 0 || time.Until(recoveryDeadline) < 0 {
+				return
 			}
-		}
-
-		// no time left, report error
-		if time.Until(protocolDeadline) >= 0 {
-			return
-		}
-
-		switch v := err.(type) {
-		case ErrTCPHeaderLength:
-			if mb.LinkRecoveryTimeout > 0 {
+			if err == io.EOF || err == io.ErrUnexpectedEOF || err == syscall.ECONNRESET {
+				mb.logf("modbus: connection closed by remote side: %v", err)
+				res = readResultCloseRetry
+				return
+			}
+			if _, ok := err.(ErrTCPHeaderLength); ok {
 				// TCP header not OK - retry with another query
 				res = readResultRetry
 				return
 			}
-			// no time left, report error
+			// other error - report
 			return
-		case errTransactionIDMismatch:
-			// the first condition check for a normal transaction id mismatch. The second part of the condition check for a wrap-around. If a wraparound is
-			// detected (last attempt is smaller than last success), the id can be higher than the last success or lower than the last attempt, but not both
-			if (v.got > mb.lastSuccessfulTransactionID && v.got < mb.lastAttemptedTransactionID) ||
-				(mb.lastAttemptedTransactionID < mb.lastSuccessfulTransactionID && (v.got > mb.lastSuccessfulTransactionID || v.got < mb.lastAttemptedTransactionID)) {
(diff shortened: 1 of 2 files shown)


tcpclient.go
```diff
@@ -288,54 +288,50 @@ func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryD
...
+					(mb.lastAttemptedTransactionID < mb.lastSuccessfulTransactionID && (v.got > mb.lastSuccessfulTransactionID || v.got < mb.lastAttemptedTransactionID)) {
+					// most likely, we simply had a timeout for the earlier query and now read the (late) response. Ignore it
+					// and assume that the response will come *without* sending another query. (If we send another query
+					// with transactionId X+1 here, we would again get a transactionMismatchError if the response to
+					// transactionId X is already in the buffer).
+					continue
+				}
 				res = readResultRetry
 				return
 			}
```
sansmoraxz, line 329: FYI nothing was done previously for unitId / protocolId mismatch

sansmoraxz:
cc @mkrafeld, please check if this mitigates the issue. Added guards for the array length check too. at least it shouldn't panic anymore

mkrafeld:
> cc @mkrafeld, please check if this mitigates the issue. Added guards for the array length check too. at least it shouldn't panic anymore

I just had a first look at it but that should fix the panic. But I would also like to get a review from someone with a bit more knowledge like @antoineco  😄

tcpclient.go
```diff
@@ -370,7 +370,30 @@ func (e errTransactionIDMismatch) Error() string {
...
+	got, expected byte
+}
+
+func (e errUnitIDMismatch) Error() string {
+	return fmt.Sprintf("modbus: response unit id '%v' does not match request '%v'", e.got, e.expected)
+}
+
 func verify(aduRequest []byte, aduResponse []byte) (err error) {
+	// len guard check for conversion
+	if len(aduRequest) < 2 {
```
antoineco: Why was the value '2' chosen here?

If a request or response must contain at least 2 bytes, I recommend adding a constant (either local to the function or globally) with a name describing the reason for using '2', such as `minAduLength`.

Otherwise, comparing to `len == 0` seems clearer.

tcpclient_test.go
```diff
@@ -165,7 +165,7 @@ func TestTCPTransactionMismatchRetry(t *testing.T) {
 	}()
 	handler := NewTCPClientHandler(ln.Addr().String())
 	handler.Timeout = 1 * time.Second
-	handler.ProtocolRecoveryTimeout = 50 * time.Millisecond
+	handler.ProtocolRecoveryTimeout = 500 * time.Millisecond
```
antoineco: A comment would probably help understand why this value was chosen, taking into account possible timeout and retries.

antoineco (commented):
The code looks good to me at first glance but wasn't tested in prod.

tcpclient.go
sansmoraxz: Cause that's what triggered the panic.

Inside `binary.BigEndian.Uint16`

```go
// Uint16 returns the uint16 representation of b[0:2].
func (bigEndian) Uint16(b []byte) uint16 {
	_ = b[1] // bounds check hint to compiler; see golang.org/issue/14808
	return uint16(b[1]) | uint16(b[0])<<8
}
```

Bounds check for casting should cover the target length rather than just emptiness check.

sansmoraxz: Looking more at the code base, this guard exists for the rtu over udp too. Maybe it should exist for every method?

tcpclient_test.go
sansmoraxz: Seemed a bit flakey on my machine. Sometimes the deadline time was over before the recovery code could execute.

sansmoraxz (commented):
resolving

tcpclient_test.go
sansmoraxz: FYI previously `ProtocolRecoveryTimeout` was just comparision against 0.

sansmoraxz: Also maybe this is a culprit here.

```go
		// ensure that answer is only written after second read attempt failed
		time.Sleep(2500 * time.Millisecond)
```

sansmoraxz: Wait yes, that was the reason. Adding a sleep before the final read attempt and using 50ms.

sansmoraxz:
OK ready

tcpclient.go
antoineco: Ah right, it's bound to the int16 size. Thanks for pointing that out.

In this case
- I agree that we should have a check everywhere it matters
- I would rename the constant to `sizeInt16` for clarity and precision.

sansmoraxz: done

antoineco (approved):

PR #123, opened 2025-12-04
feat: introduce serial link recovery

### LINKED ISSUE #122: feature: Auto recovery for serial connections
Automatically attempt to recover serial connections if gone out of sync

Like how we have with TCP with `ProtocolRecoveryTimeout` and `LinkRecoveryTimeout`

sansmoraxz:
Attempt protection against shoddy wires in RTU connection.

Fixes: #122

### DIFF
--- a/ascii_transport_test.go
+++ b/ascii_transport_test.go
@@ -1,0 +1,89 @@
+//go:build darwin || linux || freebsd || openbsd || netbsd
+// +build darwin linux freebsd openbsd netbsd
+
+// Copyright 2014 Quoc-Viet Nguyen. All rights reserved.
+// This software may be modified and distributed under the terms
+// of the BSD license. See the LICENSE file for details.
+
+package modbus
+
+import (
+	"bytes"
+	"context"
+	"testing"
+	"time"
+)
+
+func TestASCIISerialTransporter_Send_PTY(t *testing.T) {
+	master, slavePath, err := openPTY()
+	if err != nil {
+		t.Skipf("Skipping PTY test: %v", err)
+	}
+	defer master.Close()
+
+	// Request: 01 03 00 00 00 01 (Read Holding Registers)
+	// ASCII: :010300000001FB\r\n
+	reqASCII := []byte(":010300000001FB\r\n")
+
+	// Response: 01 03 02 00 00
+	// ASCII: :0103020000FA\r\n
+	respASCII := []byte(":0103020000FA\r\n")
+
+	transporter := &asciiSerialTransporter{}
+	transporter.Address = slavePath
+	transporter.BaudRate = 19200
+	transporter.Timeout = 1 * time.Second
+	transporter.IdleTimeout = serialIdleTimeout
+
+	// Start a goroutine to read request and write response to master
+	go func() {
+		buf := make([]byte, 1024)
+		n, err := master.Read(buf)
+		if err != nil {
+			return
+		}
+		if !bytes.Equal(buf[:n], reqASCII) {
+			// t.Errorf would be racy here, just log or ignore
+			return
+		}
+		// Write response
+		_, err = master.Write(respASCII)
+		if err != nil {
+			t.Errorf("Failed to write response: %v", err)
+		}
+	}()
+
+	ctx := context.Background()
+	aduResponse, err := transporter.Send(ctx, reqASCII)
+	if err != nil {
+		t.Fatalf("Send failed: %v", err)
+	}
+
+	if !bytes.Equal(aduResponse, respASCII) {
+		t.Errorf("Expected response %s, got %s", respASCII, aduResponse)
+	}
+}
+
+func TestASCIISerialTransporter_Timeout_PTY(t *testing.T) {
+	master, slavePath, err := openPTY()
+	if err != nil {
+		t.Skipf("Skipping PTY test: %v", err)
+	}
+	defer master.Close()
+
+	reqASCII := []byte(":010300000001FB\r\n")
+
+	transporter := &asciiSerialTransporter{}
+	transporter.Address = slavePath
+	transporter.BaudRate = 19200
+	transporter.Timeout = 100 * time.Millisecond
+	transporter.IdleTimeout = serialIdleTimeout
+
+	// Don't write anything to master
+
+	ctx := context.Background()
+	_, err = transporter.Send(ctx, reqASCII)
+	if err == nil {
+		t.Fatal("Expected timeout error, got nil")
+	}
+}
--- a/ascii_transport_test.go
+++ b/ascii_transport_test.go
@@ -10,8 +10,16 @@ package modbus
 import (
 	"bytes"
 	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"path/filepath"
+	"strings"
 	"testing"
 	"time"
+
+	serialpkg "github.com/grid-x/serial"
 )
 
 func TestASCIISerialTransporter_Send_PTY(t *testing.T) {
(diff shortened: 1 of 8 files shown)


sansmoraxz:
Not sure how to make test cases for this though.

sansmoraxz:
Hey @antoineco I observed that there's no delay between reads and writes in ASCII mode like there's for RTU:

https://github.com/grid-x/modbus/blob/5741db46a8fad6fce0975bb9d756edd5e75411dd/rtuclient.go#L287-L300

~~Is this intended behaviour or a missed bug?~~

EDIT: As per protocol not really necessary but recomended.

serial.go
```diff
@@ -25,8 +25,13 @@ type serialPort struct {
...
 	serial.Config
 
-	Logger      Logger
+	Logger Logger
+	// IdleTimeout is the duration to close the connection when no activity.
 	IdleTimeout time.Duration
+	// Silent period after successful connection
+	ConnectDelay time.Duration
+	// Recovery timeout if the connection is lost
+	LinkRecoveryTimeout time.Duration
```
alexjoedt, line 37: I noticed that neither `NewRTUClientHandler` nor `NewASCIIClientHandler` set a default for `LinkRecoveryTimeout`. With a zero value, `linkRecoveryDeadline` is immediately in the past, so the recovery loop will never actually retry, it will always hit "link recovery timeout reached" on the first error.

The TCP transporter handles this by checking `if mb.LinkRecoveryTimeout == 0` before giving up. Should we either set a sensible default here or add the same zero-check in the serial Send methods?

rtuclient.go
```diff
@@ -264,24 +264,72 @@ func (mb *rtuSerialTransporter) Send(ctx context.Context, aduRequest []byte) (ad
...
-	mb.logf("modbus: send % x\n", aduRequest)
-	if _, err = mb.port.Write(aduRequest); err != nil {
+	connDeadline := time.Now().Add(mb.Timeout)
+	linkRecoveryDeadline := time.Now().Add(mb.LinkRecoveryTimeout)
+
+	for {
+		// Send the request
+		mb.logf("modbus: send % x\n", aduRequest)
+		if _, err = mb.port.Write(aduRequest); err != nil {
+			if err == io.EOF || err == io.ErrUnexpectedEOF || err == syscall.ECONNRESET {
```
alexjoedt: The reconnect-on-error block is repeated four times across `rtuSerialTransporter.Send` and `asciiSerialTransporter.Send` (write error + read error in each). Would it make sense to extract something like a `shouldRecover(err) bool` helper and/or a `reconnect(ctx) error` method on `serialPort`? That way future changes only need to happen in one place.

alexjoedt: Could we use `errors.Is()` instead of `==` here? Something like `errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.ECONNRESET)`. That way we'd also catch wrapped errors. Same in asciiclient.go:193 and asciiclient.go:218 (see comment above - could be solved with a helper method).

rtuclient.go
```diff
@@ -264,24 +264,72 @@ func (mb *rtuSerialTransporter) Send(ctx context.Context, aduRequest []byte) (ad
...
+		// functionFail := aduRequest[1] & 0x80
+		bytesToRead := calculateResponseLength(aduRequest)
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(mb.calculateDelay(len(aduRequest) + bytesToRead)):
+		}
+
+		aduResponse, err = readIncrementally(aduRequest[0], aduRequest[1], mb.port, connDeadline)
+		mb.logf("modbus: recv % x\n", aduResponse[:])
```
alexjoedt: If `readIncrementally` returns `nil` for `aduResponse`, this line will panic since `aduResponse[:]` is called before the error check. Could we move this log line after the `if err != nil` block, or guard it with if `aduResponse != nil`?

rtuclient.go
```diff
@@ -264,24 +264,72 @@ func (mb *rtuSerialTransporter) Send(ctx context.Context, aduRequest []byte) (ad
 	mb.lastActivity = time.Now()
 	mb.startCloseTimer()
 
-	// Send the request
-	mb.logf("modbus: send % x\n", aduRequest)
-	if _, err = mb.port.Write(aduRequest); err != nil {
+	connDeadline := time.Now().Add(mb.Timeout)
+	linkRecoveryDeadline := time.Now().Add(mb.LinkRecoveryTimeout)
+
+	for {
```
alexjoedt, line 269: Just thinking out loud: `if close()` + `connect()` keeps succeeding but the remote device stays unresponsive, this could spin pretty fast until `linkRecoveryDeadline`. Would it be worth adding a small backoff delay between reconnect attempts, or capping the number of retries?

rtuclient.go
```diff
@@ -168,7 +168,7 @@ func readIncrementally(slaveID, functionCode byte, r io.Reader, deadline time.Ti
 
 	for {
 		if time.Now().After(deadline) { // Possible that serialport may spew data
-			return nil, errors.New("failed to read from serial port within deadline")
+			return nil, context.DeadlineExceeded
```
alexjoedt: This changed from a custom error string to `context.DeadlineExceeded`. Since that sentinel usually means the context deadline was exceeded, callers might misinterpret this as a context cancellation rather than an internal serial read timeout. Maybe wrap it to keep the distinction, like `fmt.Errorf("failed to read from serial port within deadline: %w", context.DeadlineExceeded)`?

rtu_transport_test.go
```diff
@@ -0,0 +1,117 @@
+//go:build darwin || linux || freebsd || openbsd || netbsd
+// +build darwin linux freebsd openbsd netbsd
```
alexjoedt: The build tags include darwin but `openPTY()` uses `syscall.TIOCSPTLCK` and `syscall.TIOCGPTN`, which are Linux-only. This won't compile on macOS. Should we either restrict the tag to linux or use something like `github.com/creack/pty` for cross-platform PTY support?

rtu_transport_test.go
```diff
@@ -0,0 +1,117 @@
...
+	if errno != 0 {
+		master.Close()
+		return nil, "", errno
+	}
+
+	slavePath = fmt.Sprintf("/dev/pts/%d", ptyno)
+	return master, slavePath, nil
+}
+
+func TestRTUSerialTransporter_Send_PTY(t *testing.T) {
```
alexjoedt, line 99: Nice to see PTY-based tests! 
Any plans to also add a test for the recovery path? e.g. simulating an `EOF` mid-communication and verifying that the reconnect logic kicks in. 
Right now the tests only cover the happy path and a plain timeout.

alexjoedt (commented):
Hey, sorry for the late review! Really like the direction here.

I left a few comments, mostly around defaults and some small robustness things. Hope they're helpful, let me know what you think!

serial.go
sansmoraxz: better to be same for consistency IMO

sansmoraxz: But really if went with default what would be sensible here? It would depend on parameters like bauderate, voltage stablity etc. Also if there's multiple requests in a loop line (with cetain devices turned off) would delay the entire request response cycle.

rtu_transport_test.go
sansmoraxz: For now I am dropping mac support. I don't have any mac device to test this with anyway. If required perhaps someone else can take a dig at this later on.

rtu_transport_test.go
sansmoraxz: done. + added some more edge cases.

sansmoraxz:
Also this has the unexpected surprise of writing to currently open terminals.

https://github.com/grid-x/modbus/blob/79617db65f67b431901492846a2cb78b302a98e5/Makefile#L12-L16

IMO instead of pts terminals perhaps better to have some tmp file that's auto trapped and deleted once tests are run.

rtuclient.go
sansmoraxz: True. Speaking of which wouldn't this apply to TCP too?

Perhaps it can be tackled as a seperate PR

rtuclient.go
```diff
@@ -264,24 +263,50 @@ func (mb *rtuSerialTransporter) Send(ctx context.Context, aduRequest []byte) (ad
 	mb.lastActivity = time.Now()
 	mb.startCloseTimer()
 
-	// Send the request
-	mb.logf("modbus: send % x\n", aduRequest)
-	if _, err = mb.port.Write(aduRequest); err != nil {
+	connDeadline := time.Now().Add(mb.Timeout)
```
alexjoedt: The `connDeadline` is captured once before the retry loop. After a successful `reconnect()` the remaining time on that deadline may be close to zero, causing `readIncrementally` (or `readASCII`) to time out immediately on the very next iteration and return `context.DeadlineExceeded`, which `shouldRecover` won't handle. This effectively makes link recovery useless under any real latency.

Please move `connDeadline := time.Now().Add(mb.Timeout)` inside the `for` loop (or reset it after a succesful reconnect), consistent with how `tcpclient.go` refreshes the deadline via `SetDeadline` on each iteration. Same issue exists in both `rtuclient.go` and `asciiclient.go`

serial.go
```diff
@@ -81,6 +93,28 @@ func (mb *serialPort) logf(format string, v ...interface{}) {
 	}
 }
 
+func (mb *serialPort) shouldRecover(err error) bool {
+	return errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.ECONNRESET)
```
alexjoedt: `syscall.ECONNRESET` is a TCP-layer error and won't be returned by a serial port device, right? 🤔

asciiclient.go
```diff
@@ -181,17 +182,52 @@ func (mb *asciiSerialTransporter) Send(ctx context.Context, aduRequest []byte) (
...
+					return
+				}
+				continue
+			}
+
+			return
+		}
+		// Get the response
+		aduResponse, err = readASCII(mb.port, connDeadline)
+		mb.logf("modbus: recv % x\n", aduResponse)
```
alexjoedt: Does it make sense to also add here a `nil` guard for `aduResponse` ? Like in the RTU path?

sansmoraxz: ohh forgot I missed this

serial.go
sansmoraxz: good point

rtuclient.go
sansmoraxz: Wait wouldn't this apply to `linkRecoveryDeadline` too?

sansmoraxz: Or there's a risk of infinite block

sansmoraxz: Yeah deadlock seems likely with incremental `linkRecoveryDeadline`.

We should probably mention somewhere that timeout is per read/write attempt instead of across the entire cycle as I was previously assuming.

The Timeout is coming from serial.

https://github.com/grid-x/serial/blob/ad4f461b8ed5860433b53666ddd6514cc83e0a95/serial.go#L30-L31

rtuclient.go
alexjoedt: Agreed with deferring to a follow-up PR, but please open a tracking issue so it is not
lost

ascii_transport_test.go
```diff
@@ -0,0 +1,89 @@
+//go:build darwin || linux || freebsd || openbsd || netbsd
```
alexjoedt: We should also exclude `darwin` here as it calls [openPTY()](https://github.com/grid-x/modbus/pull/123/changes#diff-54b6c2b6f171825acfc8daf92128fca5d23a5e976f64066a74b95d9b2ba1b282R18)

What do you think about to exclude is also in [asciiclient.go](https://github.com/grid-x/modbus/pull/123/changes#diff-4720b5340ad112c3d65af894b65866390c4fd9eda8e4b43c949044b90a15f833)?

serial.go
```diff
@@ -52,9 +58,14 @@ func (mb *serialPort) connect(ctx context.Context) error {
 	if mb.port == nil {
 		port, err := serial.Open(&mb.Config)
 		if err != nil {
-			return fmt.Errorf("could not open %s: %w", mb.Config.Address, err)
+			return fmt.Errorf("could not open %s: %w", mb.Address, err)
 		}
 		mb.port = port
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
alexjoedt, line 72: What about to close the serial port in case of `ctx.Done()`?

```go
select {
case <-ctx.Done():
    _ = port.Close()
    mb.port = nil
    return ctx.Err()
case <-time.After(mb.ConnectDelay):
}
```

serial.go
```diff
@@ -81,6 +92,28 @@ func (mb *serialPort) logf(format string, v ...interface{}) {
...
+
+	mb.logf("modbus: connection reset, reconnecting")
+	if cerr := mb.close(); cerr != nil {
+		mb.logf("modbus: error closing connection: %v", cerr)
+		return cerr
+	}
+	if cerr := mb.connect(ctx); cerr != nil {
+		mb.logf("modbus: error reconnecting: %v", cerr)
+		return cerr
+	}
```
alexjoedt, line 137: When a USB-serial adapter is physically unplugged mid-transfer, `port.Close()` can
itself return an error (e.g. `ENODEV` on Linux). The current code:

```go
if cerr := mb.close(); cerr != nil {
    mb.logf("modbus: error closing connection: %v", cerr)
    return cerr   // <-- bails here
}
```

…returns immediately without calling `connect()`. Note however that `close()` **does**
set `mb.port = nil` even on error, so a subsequent `connect()` call would attempt to
reopen the device — exactly what recovery is supposed to do.

Returning the close error defeats the whole purpose of link recovery in the most common
real-world scenario (hot-unplug of USB-serial dongles).

What do you think to log the close error and continue to connect?

```go
if cerr := mb.close(); cerr != nil {
    mb.logf("modbus: error closing connection: %v", cerr)
    // mb.port is nil after close(); still attempt to reopen.
}
if cerr := mb.connect(ctx); cerr != nil {
    mb.logf("modbus: error reconnecting: %v", cerr)
    return cerr
}
```

rtuclient.go
```diff
@@ -264,24 +263,50 @@ func (mb *rtuSerialTransporter) Send(ctx context.Context, aduRequest []byte) (ad
...
+				if err = mb.reconnect(ctx, err, linkRecoveryDeadline); err != nil {
+					return
+				}
+				continue
+			}
+
+			return
+		}
+		// function := aduRequest[1]
+		// functionFail := aduRequest[1] & 0x80
```
alexjoedt: These two lines are left-over from before the refactor and are never executed:
```go
// function := aduRequest[1]
// functionFail := aduRequest[1] & 0x80
```
Please remove them. If they document intent, a one-line prose comment is clearer.

asciiclient.go
```diff
@@ -181,17 +182,54 @@ func (mb *asciiSerialTransporter) Send(ctx context.Context, aduRequest []byte) (
...
+func readASCII(r io.Reader, deadline time.Time) ([]byte, error) {
 	var n, length int
 	var data [asciiMaxSize]byte
+	var err error
+
 	for {
-		if n, err = mb.port.Read(data[length:]); err != nil {
-			return
+		if time.Now().After(deadline) {
+			return nil, context.DeadlineExceeded
```
alexjoedt: `readASCII` returns the bare sentinel:
```go
return nil, context.DeadlineExceeded
```

But the equivalent path in `readIncrementally` (`rtuclient.go:172`) wraps it:
```go
return nil, fmt.Errorf("failed to read from serial port within deadline: %w", context.DeadlineExceeded)
```

Both are unwrappable with `errors.Is(err, context.DeadlineExceeded)`, but the bare form
gives callers no context about *where* or *why* the deadline was exceeded, which makes
log triage harder. Please wrap consistently:
```go
return nil, fmt.Errorf("ascii: failed to read response within deadline: %w", context.DeadlineExceeded)

File: rtuclient.go (hunk 8 above)
Comment (sansmoraxz) [reply rtuclient.go]: @alexjoedt where should it be documented?

File: rtu_transport_test.go [hunk 16]
```diff
@@ -0,0 +1,305 @@
...
+	_, err := transporter.Send(context.Background(), req)
+	if !errors.Is(err, serialpkg.ErrTimeout) {
+		t.Fatalf("expected timeout after partial response, got %v", err)
+	}
+	if got := port.written.Bytes(); !bytes.Equal(got, req) {
+		t.Fatalf("expected request %x, got %x", req, got)
+	}
+}
+
+func TestRTUSerialTransporter_ReconnectBudgetExhaustedOnReadEOF(t *testing.T) {
```
alexjoedt: The test name `TestRTUSerialTransporter_ReconnectBudgetExhaustedOnReadEOF` implies the
recovery budget ran out over time, but the test actually sets `LinkRecoveryTimeout = 0`,
which **disables** link recovery entirely (it never starts a retry loop). The error
message "link recovery timeout reached" is returned immediately on the first EOF.

A more accurate name: `TestRTUSerialTransporter_RecoveryDisabledOnReadEOF`.

Separately, it would be valuable to add a companion test with a positive
`LinkRecoveryTimeout` where the reconnect address is unreachable and confirm the error
is returned only after the budget window expires, that would be the true _budget
exhausted_ case.

rtuclient.go
sansmoraxz: Created #127

ascii_transport_test.go
alexjoedt, line 1: `rtu_transport_test.go` has three recovery-path tests:
- `TestRTUSerialTransporter_ReconnectOnMidCommunicationEOF_PTY`
- `TestRTUSerialTransporter_ReconnectBudgetExhaustedOnReadEOF`
- `TestRTUSerialTransporter_ReconnectOnWriteEOF`

`ascii_transport_test.go` covers only the happy path and a plain timeout. Given that
the ASCII recovery logic in `asciiclient.go` is symmetric with the RTU path, equivalent
tests are needed to give the same confidence. At minimum a `scriptedPort`-based test
that injects `io.EOF` on read and verifies `reconnect()` is called (or that the error
returned is "link recovery timeout reached") would close this gap.

serial.go
alexjoedt: Agree!

rtuclient.go
alexjoedt: On the [field](https://github.com/grid-x/serial/blob/ad4f461b8ed5860433b53666ddd6514cc83e0a95/serial.go#L30-L31) itself?

serial.go
sansmoraxz: I feel this would break `IdleTimeout`

sansmoraxz: Anyway adding a note that caller must close the connection if they call this. (Internal API anyway) Already handled by RTU and ASCII

serial.go
sansmoraxz: > Returning the close error defeats the whole purpose of link recovery in the most common
real-world scenario (hot-unplug of USB-serial dongles).

Good point. Although I am not familiar with any program that does this. Most often they treat it as final error state.

I am a bit confused here, let's say close failed.

```go
if cerr := mb.close(); cerr != nil {
    mb.logf("modbus: error closing connection: %v", cerr)
    // mb.port is nil after close(); still attempt to reopen.
}
if cerr := mb.connect(ctx); cerr != nil { // <- may fail due to timing
    mb.logf("modbus: error reconnecting: %v", cerr)
    return cerr
}
```

we keep spinning?

sansmoraxz: OK keeping spin, both for close and connect fails

alexjoedt (approved):
Sorry for the long review cycle, and thanks for your patience and the great contribution, @sansmoraxz!

sansmoraxz:
pinging @alexjoedt in case you approved but forgot to merge.

PR #129, opened 2026-06-14
Dynamic PTS for test runs

### LINKED ISSUE #128: Makefile tests exec: PTS files should not be hardcoded
https://github.com/grid-x/modbus/blob/79617db65f67b431901492846a2cb78b302a98e5/Makefile#L12-L16

This may give issues viz say a shell terminal is already open using that path. This would polute their std io and look whole lot of ugly.

sansmoraxz:
- Remove hard coded path which can break running cli apps (as these apaths are used by terminals). Use tmp based paths instead.
- Ensure these paths are cleaned up after test run (regardless of state)
- `glide` cli is not really necessary with modern go. So that was cleaned up too
- Moved this test into seperate script file

Fixes: #128

### DIFF
--- a/scripts/run-test-with-pty.sh
+++ b/scripts/run-test-with-pty.sh
@@ -5,9 +5,19 @@ MODE=$1
 TEST_FILTER=$2
 
 tmpdir=$(mktemp -d)
+diagslave_pids=""
+socat_pid=""
 
 cleanup() {
-    pkill -P $$ 2>/dev/null || true
+    # Kill entire process group to catch any child processes
+    if [ -n "$diagslave_pids" ]; then
+        kill $diagslave_pids 2>/dev/null || [ $? -eq 1 ]
+    fi
+    if [ -n "$socat_pid" ]; then
+        kill $socat_pid 2>/dev/null || [ $? -eq 1 ]
+    fi
+    # Fallback: kill any remaining processes in our process group
+    kill -- -$$ 2>/dev/null || [ $? -eq 1 ]
     rm -rf "$tmpdir"
 }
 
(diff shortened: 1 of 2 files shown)


scripts/run-test-with-pty.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+set -e
+
+MODE=$1
+TEST_FILTER=$2
+
+tmpdir=$(mktemp -d)
+
+cleanup() {
+    pkill -P $$ 2>/dev/null || true
```
alexjoedt: `pkill -P $$` kills only direct children of the current shell. If `diagslave` or `socat` spawns child processes, those will be orphaned. Consider using a process group kill (kill -- -$$) or tracking PIDs explicitly. 
The `|| true` suppresses all errors, including unexpected ones, which may hide issues in CI.

scripts/run-test-with-pty.sh
```diff
@@ -0,0 +1,38 @@
...
+
+case "$MODE" in
+    tcp)
+        diagslave -m tcp -p 5020 &
+        diagslave -m enc -p 5021 &
+        go test -run "$TEST_FILTER" -v .
+        ;;
+    rtu)
+        socat -d -d pty,raw,echo=0,link="$tmpdir/pty0" pty,raw,echo=0,link="$tmpdir/pty1" &
+        sleep 0.5
```
alexjoedt: `sleep 0.5` is a race condition
The `sleep 0.5 `waiting for `socat` to create the PTY symlinks is a classic timing hack. On a loaded CI machine this can be too short. A more robust approach would be to poll until the symlink appears, e.g. `until [ -e "$tmpdir/pty0" ]; do sleep 0.05; done` with a timeout guard. 
Also in [line 30](https://github.com/grid-x/modbus/pull/129/changes#diff-594c7f05d5bde5b6d1456d1ea06ba33554bb986c28ae740e7ddf334cbaaf9ab4R30)

scripts/run-test-with-pty.sh
```diff
@@ -0,0 +1,38 @@
...
+        diagslave -m rtu "$tmpdir/pty1" &
+        go test -run "$TEST_FILTER" -v .
+        ;;
+    ascii)
```
alexjoedt: `rtu` and `ascii` do the same thing except for the `-m` flag. We could avoid the copy-paste by using one shared branch/helper and just changing the MODE passed to `diagslave`.

scripts/run-test-with-pty.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+set -e
+
+MODE=$1
+TEST_FILTER=$2
```
alexjoedt, line 5: `MODE` and `TEST_FILTER` are used without much validation. `MODE` already falls through to the usage error if it’s missing, but an empty `TEST_FILTER` becomes `-run ""`, which runs all tests. If that’s intended, maybe we should document it.

scripts/run-test-with-pty.sh
alexjoedt, line 1: `set -e` won’t catch failures from background processes. If `diagslave` or `socat` fails to start, the later go test may fail with a less clear error. It would be easier to debug if we check that the background process is actually running before continuing. Wdyt?

scripts/run-test-with-pty.sh
sansmoraxz: They don't spawn any but still good point.

scripts/run-test-with-pty.sh
sansmoraxz: Validation already existed but I think the cleanup script would still be trapped. Anyway clearing up.

scripts/run-test-with-pty.sh
sansmoraxz: Not process specifically, the diaslave logs are fairly obvious to debug (it prints everything into stdout). Already scanning for the ports for be live, which IMO is a far better check instead of process liveliness.

alexjoedt (approved):
LGTM!

PR #132, opened 2026-07-16
Improve build pipelines and code quality

### LINKED ISSUE #131: [Tooling] Replace golint with modern alternative
[golint](https://github.com/golang/lint) used in the CI pipelines is no longer maintained. Most modern go apps tend to prefer using [staticcheck](https://staticcheck.dev/) or [golangci-lint](https://golangci-lint.run/)

At least for the cases not covered by the formatter.

There's also the [revive linter](https://golangci-lint.run/docs/linters/configuration/#revive) for golangci-lint as a drop in replacement for rules of golint.

sansmoraxz:
Fixes: #131

Added `golangci-lint` with a couple of rules that made sense.

And their related fixes.

  - `revive`: the defacto golint replacement
  - `govet`: just simplified `go vet`
  - `gosec`: Scan for potential security issues. Did catch a couple of casting related bugs in code.
  - `gocritic`: highly configurable. did catch the exit and defer skip in cli for err case and some comment spacing.
  - `staticcheck`: https://staticcheck.dev/
  - `errcheck`: catch where you forgot to handle `err` returned
  - `dupl`: Code deduplication scanner.

You can check other rules here: https://golangci-lint.run/docs/linters/configuration.

Updated makefile to use `golangci-lint` instead of `golint`.

Also CI pipeline has really old versions of golangci-lint.

```sh
docker run --rm public.ecr.aws/gridx/base-images:modbus-dev-1.21.latest golangci-lint --version
golangci-lint has version 1.56.1 built with go1.22.0 from a25592b5 on 2024-02-08T18:03:33Z
```

The provided `.golangci.yml` would not be accepted by that version. Ensure to update before merging.

Maintainers already granted write access to this branch.


crc.go
```diff
@@ -4,69 +4,60 @@
...
+	0x4400, 0x84C1, 0x8581, 0x4540, 0x8701, 0x47C0, 0x4680, 0x8641,
+	0x8201, 0x42C0, 0x4380, 0x8341, 0x4100, 0x81C1, 0x8081, 0x4040,
 }
 
 // Cyclical Redundancy Checking
 type crc struct {
-	high byte
-	low  byte
+	value16 uint16
 }
```
sansmoraxz, line 45: This got flagged by dupl. Used codex to refactor this.

Also, this may give a slightly better perf (if at all measureable) due to cache vectorization.

cmd/modbus-cli/main.go
```diff
@@ -71,27 +75,37 @@ func main() {
 
 	if len(os.Args) == 1 {
 		flag.PrintDefaults()
-		return
+		return 0
```
leonklingele:
```suggestion
		return 2
```

Please exit with `2` as that's what Go (and other languages) use as a de-facto standard for arg errors: https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/flag/flag.go;l=380

rtuclient.go
```diff
@@ -99,7 +99,7 @@ func (mb *rtuPackager) Encode(pdu *ProtocolDataUnit) (adu []byte, err error) {
 	checksum := crc.value()
 
 	adu[length-1] = byte(checksum >> 8)
-	adu[length-2] = byte(checksum)
+	adu[length-2] = byte(checksum & 0x00ff)
```
leonklingele, line 102: To other reviewers wondering why this change was made:

```
Why: To make the developer's intent explicitly clear and visually symmetrical.

In Modbus RTU, the 16-bit CRC (Cyclic Redundancy Check) is appended to the end of the packet, sending the lower byte first, followed by the upper byte.

In Go, converting a 16-bit integer to a byte using byte(checksum) automatically drops the higher 8 bits and keeps the lower 8 bits. It works perfectly fine on its own.

However, by adding the bitwise AND mask (& 0x00ff), the code explicitly states: "I am intentionally extracting the lower 8 bits here."

It also provides nice visual symmetry with the line immediately above it (byte(checksum >> 8)), showing exactly how the 16 bits are being split into two 8-bit chunks. Additionally, explicit masking like this often pacifies strict static analysis tools (linters) that might warn about implicit data truncation.
```

Thanks @sansmoraxz! 😊

crc.go
leonklingele, line 1: TODO @leonklingele: Review this.

tcpclient.go
```diff
@@ -115,11 +115,17 @@ func (mb *tcpPackager) Encode(pdu *ProtocolDataUnit) (adu []byte, err error) {
...
 	binary.BigEndian.PutUint16(adu[2:], tcpProtocolIdentifier)
 	// Length = sizeof(SlaveID) + sizeof(FunctionCode) + Data
-	length := uint16(1 + 1 + len(pdu.Data))
+	if len(pdu.Data) > tcpMaxLength-tcpHeaderSize-1 {
+		return nil, fmt.Errorf("modbus: length of data '%v' must not be bigger than '%v'", len(pdu.Data), tcpMaxLength-tcpHeaderSize-1)
+	}
+	length := uint16(2)
+	for range pdu.Data {
+		length++
+	}
```
leonklingele: Why not `+= len(pdu.Data)` as before? 🤔

tcpclient.go
```diff
@@ -115,11 +115,17 @@ func (mb *tcpPackager) Encode(pdu *ProtocolDataUnit) (adu []byte, err error) {
...
 	// Transaction identifier
 	transactionID := atomic.AddUint32(&mb.transactionID, 1)
-	binary.BigEndian.PutUint16(adu, uint16(transactionID))
+	binary.BigEndian.PutUint16(adu, uint16(transactionID&0xffff))
 	// Protocol identifier
 	binary.BigEndian.PutUint16(adu[2:], tcpProtocolIdentifier)
 	// Length = sizeof(SlaveID) + sizeof(FunctionCode) + Data
-	length := uint16(1 + 1 + len(pdu.Data))
+	if len(pdu.Data) > tcpMaxLength-tcpHeaderSize-1 {
+		return nil, fmt.Errorf("modbus: length of data '%v' must not be bigger than '%v'", len(pdu.Data), tcpMaxLength-tcpHeaderSize-1)
```
leonklingele:
```suggestion
		return nil, fmt.Errorf("modbus: length of data '%d' must not be bigger than '%d'", len(pdu.Data), tcpMaxLength-tcpHeaderSize-1)
```

cmd/modbus-cli/main.go
leonklingele, line 1: TODO @leonklingele: Lots of changes, still open for review.

client.go
```diff
@@ -666,20 +615,58 @@ func dataBlock(value ...uint16) []byte {
 }
 
 // dataBlockSuffix creates a sequence of uint16 data and append the suffix plus its length.
-func dataBlockSuffix(suffix []byte, value ...uint16) []byte {
+func dataBlockSuffix(suffix []byte, value ...uint16) ([]byte, error) {
+	if len(suffix) > math.MaxUint8 {
+		return nil, fmt.Errorf("modbus: suffix length '%v' exceeds maximum '%v'", len(suffix), math.MaxUint8)
```
leonklingele:
```suggestion
		return nil, fmt.Errorf("modbus: suffix length '%d' exceeds maximum '%d'", len(suffix), math.MaxUint8)
```

Untested, but this should work? 😅

client.go
```diff
@@ -666,20 +615,58 @@ func dataBlock(value ...uint16) []byte {
...
 	data := make([]byte, length+1+len(suffix))
 	for i, v := range value {
 		binary.BigEndian.PutUint16(data[i*2:], v)
 	}
-	data[length] = uint8(len(suffix))
+	var suffixLength byte
+	for range suffix {
+		suffixLength++
+	}
+	data[length] = suffixLength
```
leonklingele: Why did you change this? 😅 Seems much more complicated than before.

client.go
leonklingele, line 1: All read funcs (and some of the write funcs) still use common / duplicate code. For consistency, I'd prefer to either duplicate code everywhere (as before), or unify it for all read + write funcs.

leonklingele (commented):
Thanks for the PR, @sansmoraxz! 😊 Always very happy to see your contributions to the repo!

Left some smaller nits but also changes which, imo, are blockers. Definitely open to discuss things though! 😃

tcpclient.go
sansmoraxz: oops! fixed!

client.go
sansmoraxz: Was missed ssry. Brainfart on my side trying to resolve sec issue without disabling lint.

sansmoraxz:
I was thinking, perhaps it would be better to have typed errors instead of string builders scattered across codebase. WDYT?

Anyway might be better to keep it out of this PR, as that's a large behaviour change.

crc.go
sansmoraxz: Validated with a seperate script. Looks correct.

PR #133, opened 2026-09-01
Stop duplicating requests, close desynchronized connections

hnicolaysen:
## The bug

The TCP transporter could put the **same Modbus frame, including the same transaction ID, on the wire twice**. For a write function code that means the device executes the command twice.

This surfaced on a Huawei Sun2000 + EMMA installation — the first setup we know of that writes to **two different slave IDs over one TCP connection**. Control worked for a while, then the device began rejecting every write with exception 4 (server device failure). A packet capture showed the duplicated frame.

Two defects compose into it:

**1. A read timeout kept the connection open unconditionally** (`Send`, `readResultDone` branch). But `io.ReadFull` for the header, and the body read in `processResponse`, can fail *after* consuming part of a frame. Retaining that connection leaves the rest of the frame queued, so the next `Send` parses those leftover bytes as a header — and the stream stays misaligned from then on. The existing comment assumed a late response could always be drained via transaction-ID matching, which only holds while frame boundaries are intact.

**2. `readResultRetry` re-wrote the identical `aduRequest` buffer.** It was reached in exactly the situations where the stream was already misaligned (`ErrTCPHeaderLength`) or where a response could not be attributed to the request. So the resend both duplicated the request and compounded the desync.

The leniency window could not break out of it either: a duplicate of an already-consumed response has `got == lastSuccessfulTransactionID`, which fails the window check and falls through to the resend. Worse, the resend *appeared to work* — it consumed the stray frame, re-asked, and returned **success** to the caller — while leaving a second answer to the same transaction ID in flight. That answer fails the window check on the next request, which resends again. This is why the field symptom was permanent rather than intermittent: the recovery mechanism was manufacturing the condition it appeared to fix.

## Evidence

Both defects, reproduced against `master` before the fix:

```
conn must be nil after a mid-frame timeout, got &{{0x40000dc480}}

call returned err=<nil>
expected exactly 1 request on the wire, got 2: [1 1]
```

Transaction ID `1` written twice — and the caller got `err=<nil>`, so a duplicate write was reported to the application as a success.

## The change

- Add `ErrStreamDesynced` and mark the mid-frame cases with it: a header read that consumed `n > 0` bytes, and every `processResponse` failure (the header was already fully consumed there, so it is mid-frame by construction).
- `Send` now keeps the connection only on a **clean** timeout — nothing consumed, stream still frame-aligned. A mid-frame timeout closes.
- Remove `readResultRetry`. The transporter cannot safely retry: it does not know whether the request was a write, and it cannot re-encode with a fresh transaction ID because it does not own the packager. That decision belongs to the caller, and reissuing there encodes a new transaction ID.
- Extract the out-of-window predicate into `isLateResponse` — the wraparound logic is preserved verbatim, just named and readable.

### Deliberately unchanged

- **The in-window "late response" drain.** Discarding a stale response from an *intact* stream without resending is correct, and it is the one case where the connection really is healthy. `TestTCPTransactionMismatchRetry` covers this and still passes untouched.
- **`readResultCloseRetry`** for link recovery, where the remote closed before sending anything at all — there the stream is empty rather than misaligned. Note it does reissue the request after reconnecting, which is a narrower instance of the same class of problem; left alone here to keep link-recovery semantics (including the recently added serial path) untouched.

## Tests

Two additions to `tcpclient_test.go`, both failing before this change and passing after:

- `TestTCPMidFrameTimeoutClosesConnection` — a server that writes a partial header and stalls must leave `conn == nil`.
- `TestTCPNeverResendsSameTransactionID` — a server answering with an unattributable transaction ID must never see the same request twice on the wire.

`make ci_test` passes in full: TCP, RTU and ASCII suites against diagslave and PTYs. The `TCP` filter picks up both new tests.

## Conformance

Checked against *MODBUS Messaging on TCP/IP Implementation Guide V1.0b*.

Two independent rules are at work here, and they are easy to conflate:

**Protocol legality.** §4.4.1.2: "at a time, **on a TCP connection**, this identifier must be unique." Re-writing the same ADU on the *same* connection while the original transaction is still pending creates two live transactions carrying one identifier, so the old behaviour was a protocol violation and removing it is a conformance fix. Note the scoping: reissuing a request on a freshly dialled connection does **not** violate this, because the old socket — and with it the pending transaction — is gone.

**Idempotency.** Whether executing a request twice changes the device. The spec says nothing about this; it is a property of physical equipment. Reads are free to repeat, writes are not.

Each rule is applied uniformly to reads and writes:

| Path | Governed by | Before | After |
|---|---|---|---|
| Same-connection resend on protocol desync | §4.4.1.2 — illegal | resent the identical ADU, sometimes returning success | closes and reports `ErrStreamDesynced` — reads *and* writes |
| Fresh-connection reissue on link recovery | idempotency — spec-legal either way | reissued any request | reads reissued as before, writes reported |

So the function-code gate is **not** a spec requirement. It rests on idempotency alone. (RTU has no transaction identifier at all, which is why the same hazard on the serial path — see below — is purely an idempotency problem.)

Supporting the layering: Figure 11 routes a retry back through *Build a MODBUS Request* — a fresh transaction, not a byte-identical replay — and §4.4.1.4 warns that retrying faster than the maximum reasonable response time risks "excessive congestion at the target device", which "should always be avoided".

One conscious deviation: §4.4.1.3 says a response whose transaction identifier matches no pending transaction "must be discarded", i.e. discard it and keep reading, without implicating the connection. This change closes instead. That is stricter than the standard; it is deliberate, because the same branch is also reachable when the stream really is misaligned (the tail of an earlier frame can parse as a plausible header) and the two cases cannot be distinguished from the outside.

## Behaviour changes for consumers

The API is additive — `ErrStreamDesynced` is new, and `readResultRetry` and `flush` were unexported — so nothing fails to compile. Everything below is behavioural.

**Writes.** Every path that reissued a write is now closed, on both rules above. A caller that previously saw a write succeed after a link hiccup will now get an error. This is not optional: any way of preserving it re-executes a command on live equipment, which is the defect this PR exists to fix.

**Reads**, versus the base commit:

| Situation | Before | After | |
|---|---|---|---|
| Header read fails with EOF/ECONNRESET | reconnect and reissue | unchanged | |
| Header read times out mid-frame | connection kept (desynchronised), error returned | connection closed, error returned | caller sees an error either way |
| `processResponse` hits EOF/ECONNRESET mid-body | reconnect and reissue | unchanged | |
| `processResponse` reports an implausible length | same-connection resend, could succeed | hard error | **regression** |
| `verify` reports an out-of-window transaction ID | same-connection resend, could succeed | hard error | intended — see *The bug* |

The out-of-window case is the loop that made the field symptom permanent, so failing there is the point. The implausible-length case is the one remaining collateral change: there the stream really is misaligned, so the old same-connection resend was unsafe — it could be routed through link recovery instead (reconnect, then reissue if repeatable), but that is a broader claim about reconnecting being the right answer to a device emitting garbage, so it is left failing. Say so in review if you disagree.

Everything else on the read path is unchanged, and the rule is uniform: **socket gone with nothing queued** means link recovery may redial and reissue a repeatable request; anything else means desync, close and report.

Also worth knowing: this fixes the TCP transport only. `rtuclient.go` and `asciiclient.go` still reissue writes verbatim after reconnecting — same hazard, tracked in #135 (the repeatability policy lives in `funcCodeRepeatable` in `modbus.go` precisely so that fix is small). And link recovery has no retry interval at all: #134.

One property this PR does not change, worth stating so it is not mistaken for a guarantee: a write that fails *after* being sent is inherently in doubt. The transport no longer reissues it, but it also cannot tell the caller whether the device acted on it, so an upper layer with a generic retry may still send it a second time — once, with a fresh transaction ID, rather than the unbounded identical-ADU storm this PR removes. Most register writes are idempotent in effect, so this is usually harmless; for command-style registers it is a decision for the caller.

## Notes for reviewers

- `ErrStreamDesynced` is the only added API surface. It is what lets a caller tell a desync apart from a plain timeout and retry deliberately — the protocol error types stay unexported. Happy to export those separately if useful.
- On how much the read regressions above actually bite: `LinkRecoveryTimeout` is measured from the start of `Send` and is typically far shorter than a device's response time (in our fleet the median read is ~158 ms against a 53 ms budget, so ~85% of reads could never reach that recovery path anyway). Callers with their own reconnect-and-retry are unaffected. And removing the silent in-`Send` reconnect is itself an improvement: a connection replaced invisibly skips any post-connect initialisation the caller performs.

Refs: https://grid-x.atlassian.net/browse/INTE-6086
Related: #56 (same-connection multi-slave-ID use, which is how we hit this)

### DIFF
--- a/modbus.go
+++ b/modbus.go
@@ -57,9 +57,14 @@ const (
 // Reference) can write. Deciding that needs the byte after the function code,
 // which is past what this predicate looks at.
 //
-// Transport-agnostic on purpose - the function code sits at a different offset in
-// each framing (adu[1] for RTU and ASCII, adu[tcpHeaderSize] for TCP), so each
+// Kept free of framing detail on purpose: the function code sits at a different
+// offset per framing (adu[tcpHeaderSize] for TCP, adu[1] for RTU and ASCII), so a
 // transport extracts it and this decides the policy.
+//
+// Only the TCP transport consults it today. The RTU and ASCII transports still
+// reissue any request verbatim after reconnecting, so the same duplicate-write
+// hazard remains there - tracked separately, and this predicate is placed here so
+// that fix does not need its own copy of the policy.
 func funcCodeRepeatable(fc byte) bool {
 	switch fc {
 	case FuncCodeReadCoils,
--- a/modbus.go
+++ b/modbus.go
@@ -40,6 +40,39 @@ const (
 	FuncCodeReadDeviceIdentification = 43
 )
 
+// funcCodeRepeatable reports whether a request carrying this function code may be
+// written to the wire more than once. Link recovery reissues a request verbatim
+// after reconnecting, which is harmless for a read but means a device that
+// processed a write before the connection dropped executes it again - once per
+// recovery attempt.
+//
+// Only purely reading function codes are repeatable. Everything else counts as a
+// write, including FuncCodeReadWriteMultipleRegisters (which also writes) and any
+// vendor-specific code we cannot reason about: misjudging a read costs one failed
+// request, misjudging a write duplicates a command on live equipment.
+//
+// FuncCodeReadDeviceIdentification is excluded for the same reason. It is
+// Encapsulated Interface Transport, and only MEI type
+// meiTypeReadDeviceIdentification is a read - MEI type 13 (CANopen General
+// Reference) can write. Deciding that needs the byte after the function code,
+// which is past what this predicate looks at.
+//
+// Transport-agnostic on purpose - the function code sits at a different offset in
+// each framing (adu[1] for RTU and ASCII, adu[tcpHeaderSize] for TCP), so each
+// transport extracts it and this decides the policy.
+func funcCodeRepeatable(fc byte) bool {
+	switch fc {
+	case FuncCodeReadCoils,
+		FuncCodeReadDiscreteInputs,
+		FuncCodeReadHoldingRegisters,
+		FuncCodeReadInputRegisters,
+		FuncCodeReadFIFOQueue:
+		return true
+	default:
+		return false
+	}
+}
+
 // meiType specifies a MEI Type as defined in https://www.modbus.org/docs/Modbus_Application_Protocol_V1_1b.pdf#page=44
 type meiType byte
 
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -445,14 +445,16 @@ func (mb *tcpTransporter) isLateResponse(got uint16) bool {
 
 func (mb *tcpTransporter) processResponse(data []byte) (aduResponse []byte, err error) {
 	// Read length, ignore transaction & protocol id (4 bytes)
+	// An implausible length means the bytes just read were not a frame header at
+	// all - most likely the tail of an earlier frame. There is nothing to salvage:
+	// the caller marks this as a desync and Send discards the connection, so the
+	// receive buffer goes with it and draining it here would be pointless.
 	length := int(binary.BigEndian.Uint16(data[4:]))
 	if length <= 0 {
-		mb.flush(data[:])
 		err = ErrTCPHeaderLength(length)
 		return
 	}
 	if length > (tcpMaxLength - (tcpHeaderSize - 1)) {
-		mb.flush(data[:])
 		err = ErrTCPHeaderLength(length)
 		return
 	}
@@ -570,22 +572,6 @@ func (mb *tcpTransporter) Close() error {
 	defer mb.mu.Unlock()
 
 	return mb.close()
-}
-
-// flush flushes pending data in the connection,
-// returns io.EOF if connection is closed.
-func (mb *tcpTransporter) flush(b []byte) (err error) {
-	if err = mb.conn.SetReadDeadline(time.Now()); err != nil {
-		return
-	}
-	// Timeout setting will be reset when reading
-	if _, err = mb.conn.Read(b); err != nil {
-		// Ignore timeout error
-		if netError, ok := err.(net.Error); ok && netError.Timeout() {
-			err = nil
-		}
-	}
-	return
 }
 
 func (mb *tcpTransporter) logf(format string, v ...interface{}) {
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -568,6 +568,17 @@ func (mb *tcpTransporter) close() (err error) {
 		err = mb.conn.Close()
 		mb.conn = nil
 	}
+	// Collapse the late-response window. Nothing can still be in flight on a
+	// connection that no longer exists, so no transaction ID may be treated as a
+	// late response until a new exchange succeeds on the next connection.
+	//
+	// Without this the window keeps spanning transactions that only ever existed
+	// on the previous socket, and it widens with every further attempt because
+	// lastAttemptedTransactionID advances while lastSuccessfulTransactionID stays
+	// frozen. A stray frame on a fresh connection would then be taken for a late
+	// response, silently drained, and the resulting clean timeout would keep a
+	// connection this function had just decided to discard.
+	mb.lastSuccessfulTransactionID = mb.lastAttemptedTransactionID
 	return
 }
 
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -336,6 +336,33 @@ func (mb *tcpTransporter) shouldRecover(err error) bool {
 	return errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.ECONNRESET)
 }
 
+// isRepeatable reports whether aduRequest may be written to the wire a second
+// time. Link recovery reissues the request verbatim after reconnecting, which is
+// harmless for a read but means a device that processed a write before dropping
+// the connection executes it again - once per recovery attempt.
+//
+// Only function codes that are purely read are repeatable. Everything else is
+// treated as a write, including FuncCodeReadWriteMultipleRegisters (which also
+// writes) and any vendor-specific code we cannot reason about: the cost of
+// misjudging a read is one failed request, the cost of misjudging a write is a
+// duplicated command on live equipment.
+func isRepeatable(aduRequest []byte) bool {
+	if len(aduRequest) <= tcpHeaderSize {
+		return false
+	}
+	switch aduRequest[tcpHeaderSize] {
+	case FuncCodeReadCoils,
+		FuncCodeReadDiscreteInputs,
+		FuncCodeReadHoldingRegisters,
+		FuncCodeReadInputRegisters,
+		FuncCodeReadFIFOQueue,
+		FuncCodeReadDeviceIdentification:
+		return true
+	default:
+		return false
+	}
+}
+
 func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryDeadline time.Time, protocolDeadline time.Time) (aduResponse []byte, res readResult, err error) {
 	// res is readResultDone by default, which either means we succeeded or err contains the fatal error
 	for {
@@ -351,7 +378,12 @@ func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryD
 			if mb.LinkRecoveryTimeout == 0 || time.Until(recoveryDeadline) < 0 {
 				return
 			}
-			if mb.shouldRecover(err) {
+			if mb.shouldRecover(err) && isRepeatable(aduRequest) {
+				// The stream is empty rather than misaligned, and the request is a
+				// read, so reconnecting and reissuing it has no effect on the device.
+				// A write is reported instead - it may already have been executed
+				// before the connection dropped - leaving the retry to the caller,
+				// which re-encodes with a fresh transaction ID.
 				mb.logf("modbus: connection closed by remote side: %v", err)
 				res = readResultCloseRetry
 			}
--- a/tcpclient.go
+++ b/tcpclient.go
@@ -37,6 +37,23 @@ func (length ErrTCPHeaderLength) Error() string {
 		length, tcpMaxLength-tcpHeaderSize+1)
 }
 
+// ErrStreamDesynced reports that the response stream is no longer aligned to a
+// Modbus frame boundary, so the bytes still buffered on the connection cannot be
+// attributed to any request. It is returned wrapped around the underlying cause.
+//
+// A desynchronized connection is not recoverable in place: the transporter
+// closes it so that the next Send dials afresh. Callers that want the request
+// retried must issue it again, which encodes a new transaction ID. The
+// transporter deliberately does not retry by itself — re-writing the same ADU
+// would put a duplicate frame with an identical transaction ID on the wire, and
+// for a write function code that makes the device execute the command twice.
+var ErrStreamDesynced = errors.New("modbus: response stream desynchronized")
+
+// desynced marks err as a stream desynchronization, see [ErrStreamDesynced].
+func desynced(err error) error {
+	return fmt.Errorf("%w: %w", ErrStreamDesynced, err)
+}
+
 // TCPClientHandler implements Packager and Transporter interface.
 type TCPClientHandler struct {
 	tcpPackager
@@ -197,7 +214,6 @@ type readResult int
 
 const (
 	readResultDone readResult = iota
-	readResultRetry
 	readResultCloseRetry
 )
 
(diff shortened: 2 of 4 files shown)


tcpclient.go
```diff
@@ -283,9 +316,6 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
 				mb.lastSuccessfulTransactionID = binary.BigEndian.Uint16(aduResponse)
 			}
 			return
-		case readResultRetry:
-			mb.logf("modbus: retry reading response, because of %v", err)
-			continue
 		case readResultCloseRetry:
```
hnicolaysen: The PR removes `readResultRetry`, but this path still rewrites the byte-identical `aduRequest` after reconnecting — and it does so without any bound. Probed on this branch with a server that reads a `WriteSingleRegister` and then drops the connection without answering:

```
err=modbus: read response: EOF
transaction IDs on the wire: 1 (x ~1000 within a 500ms LinkRecoveryTimeout)
unique ids: ['1']
```

(master: ~7500 — so this is pre-existing, not introduced here.)

§4.4.1.2 isn't violated, since each resend goes out on a fresh connection. But a device that processed the write before resetting gets the command executed once per iteration, which is the same field harm the PR targets, and the unbounded rate is what §4.4.1.4 warns about directly.

I'm fine with leaving it out of scope. Two things I'd ask for:

- The transporter *does* know the function code — `aduRequest[tcpHeaderSize]` — so gating this branch on read function codes is small and self-contained, and it removes the "does not know whether the request was a write" argument from the description.
- Since the diff doesn't cover it, `TestTCPNeverResendsSameTransactionID` and the PR title both claim more than the change delivers. A future reader will trust the test name over the caveat in the description. Worth narrowing both, plus a follow-up ticket for the missing backoff.

tcpclient.go
```diff
@@ -369,6 +398,19 @@ func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryD
...
+// request that we already gave up on, i.e. whether it lies strictly between the
+// last successfully verified and the currently attempted transaction ID, taking
+// uint16 wraparound into account.
+func (mb *tcpTransporter) isLateResponse(got uint16) bool {
+	if mb.lastAttemptedTransactionID < mb.lastSuccessfulTransactionID {
+		// The counter wrapped between the two, so the open interval is the union
+		// of the two ends of the uint16 range.
+		return got > mb.lastSuccessfulTransactionID || got < mb.lastAttemptedTransactionID
+	}
+	return got > mb.lastSuccessfulTransactionID && got < mb.lastAttemptedTransactionID
```
hnicolaysen, line 438: The window this reads from goes stale across a close: `close()` never resets `lastSuccessfulTransactionID`, so it keeps spanning transactions that only ever existed on the *previous* socket. Probed on this branch — send #1 desyncs and closes, then the freshly dialled connection answers with transaction ID `1`:

```
send #1 err=... ErrStreamDesynced: response transaction id '48879' does not match request '1'
send #2 err=... i/o timeout
lastSuccessful=0 lastAttempted=2 connKept=true
```

ID `1` cannot possibly be in flight on that connection, but `isLateResponse` returns true, the frame is silently drained, and the resulting clean timeout **keeps** the connection — the exact outcome this PR exists to prevent.

Pre-existing, but this PR makes it much more reachable, because closing goes from corner case to the standard reaction to a desync. Fix is one line in `close()`:

```go
mb.lastSuccessfulTransactionID = mb.lastAttemptedTransactionID
```

Only in `close()` — not on the clean-timeout keep-alive path, where an in-flight late response is legitimate and the window has to survive.

tcpclient.go
```diff
@@ -323,44 +359,37 @@ func (mb *tcpTransporter) readResponse(aduRequest []byte, data []byte, recoveryD
...
-				// TCP header not OK - retry with another query
-				res = readResultRetry
-				return
-			}
-			// other error - report
+			// The header was consumed in full, so any failure past this point
+			// leaves the stream mid-frame - whether the announced length was
+			// nonsense (which means the "header" was really the tail of an earlier
+			// frame) or the body read failed part-way through.
+			err = desynced(err)
```
hnicolaysen, line 395: Now that every `processResponse` failure lands here and `Send` closes unconditionally, the two `mb.flush(data[:])` calls in `processResponse` on the `ErrTCPHeaderLength` paths are dead weight — they drain a buffer that is about to be discarded with the connection. Harmless, but worth dropping while this reasoning is fresh.

tcpclient_test.go
```diff
@@ -292,6 +293,135 @@ func TestTCPTransactionMismatchRetry(t *testing.T) {
...
+			observed = append(observed, binary.BigEndian.Uint16(req))
+			isFirst := len(observed) == 1
+			mu.Unlock()
+
+			resp, err := packager.Encode(&ProtocolDataUnit{
+				FunctionCode: FuncCodeReadInputRegisters,
+				Data:         []byte{0x02, 0xCA, 0xFE},
+			})
+			if err != nil {
+				t.Error(err)
```
hnicolaysen: Nothing waits for this goroutine: `handler.Close()` doesn't guarantee it has left `ReadFull`, so a `t.Error` here can fire after the test function returned, which panics the run. Same shape at line 319 in the other new test. A `sync.WaitGroup`, or passing the error over a channel and asserting on it in the test body, closes it.

nelsininho (approved):

alexjoedt (commented):
Review of the desync change. Four inline comments below: two regressions against master (window collapse on reissue, ReadDeviceIdentification link recovery) and two one-line fixes in code the PR touches (ETIMEDOUT classification, ASCII offset in the doc). The core change is right and matches the Implementation Guide 4.4.1.2.

tcpclient.go
```diff
@@ -259,34 +285,13 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
...
-			}
-			return
-		case readResultRetry:
-			mb.logf("modbus: retry reading response, because of %v", err)
-			continue
-		case readResultCloseRetry:
+
+		if reconnectAndReissue {
 			mb.logf("modbus: close connection and retry reading response, because of %v", err)
 			mb.close()
```
alexjoedt, line 296: **Reissue path collapses the late-response window onto the TID it is about to reissue (regression against master, reproduced).**

`close()` now sets `lastSuccessfulTransactionID = lastAttemptedTransactionID`. This branch calls `close()` and then writes the same transaction ID X again on the fresh connection. After the call the window is `(X, X)`, which is empty, so X is no longer accepted as a late response.

Sequence that breaks:
1. Read X, header read hits EOF, `readResponse` returns `reconnectAndReissue`.
2. `close()` here sets `lastSuccessful = lastAttempted = X`. Redial, write X again.
3. The device answers slowly, the read times out cleanly, the connection is kept (correct).
4. Next `Send` with X+1 reads the late frame for X first. `isLateResponse(X)` is `X > X && X < X+1`, which is false. The frame is wrapped as `ErrStreamDesynced` and a healthy connection is closed.

On master `close()` left `lastSuccessful` at the previous verified TID, so X stayed inside the window and the request returned the device's reply. A test with exactly this sequence fails 3/3 on this branch with `response transaction id '1' does not match request '2'` and passes 3/3 on master.

Why fix: every deployment with `LinkRecoveryTimeout` and `ProtocolRecoveryTimeout` set, a device that drops idle connections, and an occasional slow answer will see spurious `ErrStreamDesynced` plus a redial where master returned data.

Recommended fix, either:
- Reopen the window for the reissued TID right after the close:
  ```go
  mb.close()
  mb.lastSuccessfulTransactionID = mb.lastAttemptedTransactionID - 1 // window is exactly {X}
  ```
  uint16 wraparound is handled by `isLateResponse` already.
- Or move the collapse out of `close()` into the terminal call sites (write error, desync, `closeIdle`, `Close`), so the reissue path never collapses. This keeps `close()` a plain close and puts the intent where the exchange really ends.

Worth adding the repro as a test next to `TestTCPCloseResetsLateResponseWindow`; happy to share it.

tcpclient.go
```diff
@@ -295,72 +300,123 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
...
 
+// isRepeatable reports whether aduRequest may be written to the wire a second
+// time, i.e. whether reissuing it during link recovery is free of side effects on
+// the device. The policy lives in funcCodeRepeatable; this only knows where MBAP
+// keeps the function code.
+func isRepeatable(aduRequest []byte) bool {
+	if len(aduRequest) <= tcpHeaderSize {
+		return false
+	}
+	return funcCodeRepeatable(aduRequest[tcpHeaderSize])
```
alexjoedt, line 342: **`ReadDeviceIdentification` lost link recovery (regression against master).**

`funcCodeRepeatable` excludes FC 43 because "deciding that needs the byte after the function code, which is past what this predicate looks at". This function has the whole ADU, so it can look at that byte. The library's only FC 43 emitter (`client.go`, `readDeviceIdentificationWithObjectID`) hardcodes `meiTypeReadDeviceIdentification` (14), which the application spec (6.21) defines as read-only. MEI 13 (CANopen, 6.20) is the one that can write.

On master a `ReadDeviceIdentification` after a device-side idle close was redialed and reissued like every other request. On this branch the first call after the idle close returns `read response: EOF` although `LinkRecoveryTimeout` is set. `TestIsRepeatable` currently pins that regression with `{"read device identification", ..., false}`.

Why fix: a first-party read method regresses for every user with `LinkRecoveryTimeout` set, and the gate has the information to decide correctly.

Recommended fix:
```go
fc := aduRequest[tcpHeaderSize]
if fc == FuncCodeReadDeviceIdentification {
    return len(aduRequest) > tcpHeaderSize+1 &&
        aduRequest[tcpHeaderSize+1] == byte(meiTypeReadDeviceIdentification)
}
return funcCodeRepeatable(fc)
```
Then flip the FC 43 row in `TestIsRepeatable` to `true` for MEI 14, add a `false` row for MEI 13, and trim the FC 43 paragraph in the `funcCodeRepeatable` doc.

tcpclient.go
```diff
@@ -295,72 +300,123 @@ func (mb *tcpTransporter) Send(ctx context.Context, aduRequest []byte) (aduRespo
...
 				continue
 			}
-		default:
-			mb.logf("modbus: unhandled read result %v", res)
-			return nil, fmt.Errorf("modbus: unhandled read result %v", res)
 		}
+
+		if err != nil {
+			var netErr net.Error
+			if errors.As(err, &netErr) && netErr.Timeout() && !errors.Is(err, ErrStreamDesynced) {
```
alexjoedt, line 307: **Kernel `ETIMEDOUT` passes this check, so a dead socket is kept as a "clean timeout".**

`syscall.Errno.Timeout()` is `e == EAGAIN || e == EWOULDBLOCK || e == ETIMEDOUT` (`syscall_unix.go`), and `net.OpError.Timeout()` unwraps `*os.SyscallError` to reach it. `ETIMEDOUT` on a read means the kernel gave up on the peer (keepalive probes exhausted or retransmit budget spent), which is the opposite of "the device may simply be slow". With `n == 0` that error is neither desynced nor in `shouldRecover`, so `Send` logs "keeping connection" for a socket the kernel already declared dead. The next `Send` writes into it, fails, and only then closes: two application errors for one outage, and a repeatable read that never enters link recovery. The default dialer leaves keepalive at Go's defaults (15s, 9 probes), so a silently vanished peer does surface this way.

This is pre-existing on master, but the diff rewrote the branch and the new comment states an invariant ("nothing of the response was consumed, so the stream is still aligned") that does not hold for `ETIMEDOUT`. Cheap to fix while the reasoning is fresh.

Recommended fix, test for our own deadline instead of the interface method:
```go
if errors.Is(err, os.ErrDeadlineExceeded) && !errors.Is(err, ErrStreamDesynced) {
```
`os.ErrDeadlineExceeded` sits directly as `OpError.Err`, so `errors.Is` reaches it, and only our deadline can fire with the socket still alive, which is exactly what the comment describes.

modbus.go
```diff
@@ -40,6 +40,44 @@ const (
...
+// request, misjudging a write duplicates a command on live equipment.
+//
+// FuncCodeReadDeviceIdentification is excluded for the same reason. It is
+// Encapsulated Interface Transport, and only MEI type
+// meiTypeReadDeviceIdentification is a read - MEI type 13 (CANopen General
+// Reference) can write. Deciding that needs the byte after the function code,
+// which is past what this predicate looks at.
+//
+// Kept free of framing detail on purpose: the function code sits at a different
+// offset per framing (adu[tcpHeaderSize] for TCP, adu[1] for RTU and ASCII), so a
```
alexjoedt, line 61: **Wrong ASCII offset in the doc.**

`adu[1]` is right for RTU (`rtuclient.go`, `adu[1] = pdu.FunctionCode`). `asciiPackager.Encode` writes `':'`, then two hex characters for the slave ID, then two hex characters for the function code, and `Decode` reads it with `readHex(adu[3:])`. In an ASCII frame `adu[1]` is the first hex digit of the slave address.

Why fix: this paragraph exists so the RTU/ASCII fix (#135) can reuse the predicate. An implementer who follows it gates ASCII on an ASCII digit (0x30..0x46), which never matches 1/2/3/4/24, so every ASCII request would count as a write and link recovery would be silently disabled for all ASCII reads.

Recommended fix: "adu[1] for RTU; for ASCII decode the two hex characters at adu[3:5] first", or drop the offset list here and let each transport document its own offset.

tcpclient.go
alexjoedt: not critical - will not be addressed

tcpclient.go
alexjoedt: not critical - will not be addressed

tcpclient.go
alexjoedt: not critical - will not be addressed

modbus.go
alexjoedt: not critical - will not be addressed

PR #136, opened 2026-09-04
ci: replace buildkite pipeline with github actions

alexjoedt:
Replaces the Buildkite pipeline with a GitHub workflow.

- New `.github/workflows/ci.yml` with three jobs: Tests, Build, Lint. Each one runs the existing `ci_test`, `ci_build` and `ci_lint` make target, so everything still runs inside the gridx dev image (Go 1.21, golint, diagslave, socat).
- Runs on pull requests and pushes to master. Runs on the same ref cancel each other.
- Anonymous pulls from public ECR hit `toomanyrequests` on the shared runners, so the image now comes from `ghcr.io/grid-x/modbus-dev:go1.21`. The new `mirror-ci-image.yml` workflow copies it there (already ran, same digest as the ECR image). It runs on changes to its own file and on manual dispatch, so bumping the default tag re-mirrors.
- Removes `.buildkite/pipeline.yaml` and the unused `BRANCH` variable in the Makefile.

All three jobs pass on this PR.

After merge:
- Disable the pipeline in Buildkite and remove the Buildkite app from this repo.
- Optionally add Tests, Build and Lint as required checks in branch protection. There are no required contexts today.

Follow-up (separate PR): drop the dev image in favour of `actions/setup-go` plus apt and a diagslave download, and replace the archived golint.


.github/workflows/mirror-ci-image.yml
leonklingele, line 1: Do we really need this file? Once the PR is merged, we will no longer publish to our ECR anyway, right?

PR #139, opened 2026-09-16
feat(modbus-cli): support function code 2 (read discrete inputs)

pj-gridx:
Wire client.ReadDiscreteInputs into the modbus-cli exec switch so that -fn-code=0x02 can be used.

FC1 and FC2 return bit-packed bytes rather than 16-bit registers, which the register-oriented parsers cannot represent: with -quantity <= 8 the raw parser printed nothing at all. Add a bit parser that prints one address per line and make it the default for both bit-wise function codes, selectable explicitly via -type-parse=bits.


README.md
```diff
@@ -137,6 +137,24 @@ Reading multiple registers is only possible in the raw format
 ./modbus-cli -address=tcp://127.0.0.1:502 -quantity=16 -type-parse=raw -register=42
 ```
 
+### Reading Coils and Discrete Inputs
+
+Function codes 0x01 (read coils) and 0x02 (read discrete inputs) return one bit per
+address instead of 16-bit registers. For these function codes `-quantity` is the number
+of coils/inputs and the result is printed as one address per line, so `-type-parse`
```
leonklingele: The arg description of `-quantity` currently says:

```
number of 16-bit registers to read/write (not bytes)
```

This should be reworded to something like:

```
for fn-codes 0x01 and 0x02: number of coils/inputs to read (one bit); for all other fn-codes: number of 16-bit registers (not bytes) to read/write
```

pj-gridx: Adapted line 55 in the readme

leonklingele (approved):

